PR #7, opened 2023-11-29
Directory layout that's more standard

vrivellino:


vince-riv-argocdk3s:
0 of 0 apps with changes as compared to manifests in [main](https://github.com/vince-riv/argo-diff/tree/main) as of _2023-11-29 15:35:30Z_

PR #34, opened 2024-01-14
feat(charts): adds a helm chart that can deploy argo-diff

knechtionscoding:
feat(docker): adds support for running as non-root user
feat(actions): adds helm chart actions. Helm lint + helm releaser


vince-riv-argocdk3s:

9:57PM UTC, 14 Jan 2024

knechtionscoding:
Combined #23 and #25  into one PR

PR #36, opened 2024-01-18
feat: compare event sha to HEAD

vrivellino:


vince-riv-argocdk3s:

9:10PM UTC, 18 Jan 2024

PR #37, opened 2024-01-23
chore/dep update

vrivellino:
- update argocd go pkg to v2.9.5
- update go-github to v58


vince-riv-argocdk3s:

2:28AM UTC, 23 Jan 2024

PR #40, opened 2024-01-24
chore: re-organize web server code

vrivellino:
- Moves bulk of cmd/main.go into internal packages
- Reduces some use of global variables as well


vince-riv-argocdk3s:

12:14AM UTC, 24 Jan 2024

PR #43, opened 2024-01-25
feat(chart): add command into the templated command and allow override

knechtionscoding:
Discovered when manually editing deployment for testing that the command isn't specific on the pod. This means if changed outside the chart then argo or flux won't revert.


vince-riv-argocdk3s:

2:08PM UTC, 25 Jan 2024

PR #45, opened 2024-01-25
fix: remove write to app-list.json (left from devel)

vrivellino:
Was used during development - shouldn't be included anymore


vince-riv-argocdk3s:

9:10PM UTC, 25 Jan 2024

PR #47, opened 2024-01-25
fix: don't comment when there are no argo changes

vrivellino:


vince-riv-argocdk3s:

9:34PM UTC, 25 Jan 2024

PR #252, opened 2026-07-06
feat: prepare for argo-diff v3

vrivellino:
- ArgoCD v3.5
- Go 1.26


vrivellino:
After reading the release notes for ArgoCD 3.5, I don't believe we need to bump the major version for argo-diff

PR #266, opened 2026-07-28
feat: match ArgoCD apps on non-github.com git hosts

nakul-detroja:
## Summary

Related to #160 (non-github.com host support — see scope note below).

`gitRepoMatch()` only recognizes `github.com` URLs (it builds `host/owner/repo` candidates in HTTP(S) and scp-style SSH forms). Applications whose source `repoURL` points at a **non-github.com host** are therefore never matched, and argo-diff produces no diff for them.

This affects, at least:
- **GitHub Enterprise Server** (`github.example.com/owner/repo.git`)
- **AWS CodeConnections** (`https://codeconnections.<region>.amazonaws.com/git-http/<account>/<region>/<uuid>/owner/repo.git`) — ArgoCD can consume these directly, and this is what surfaced the gap for us
- **GitLab / self-hosted GitLab** and other providers that preserve the `owner/repo` path
- read-only mirrors

## Change

Add a host-agnostic fallback that runs **only after** the existing github.com candidates fail to match, so current behavior is unchanged. The fallback matches the `owner/repo` path suffix (`.git` optional):

- The leading separator (`/` for HTTP(S) paths, `:` for scp-style SSH) anchors the owner boundary, so a URL ending in `…/otherowner/repo` is **not** matched for owner `owner`.
- Comparison is case-insensitive, since git hosting providers treat owner/repo names case-insensitively.

## Tests

Adds `TestGitRepoMatch` (table-driven) covering:
- github.com https (with/without `.git`) and scp-style SSH — existing behavior
- GitHub Enterprise, AWS CodeConnections, mirror without `.git`, generic scp SSH — new behavior
- case-insensitive match
- negatives: different repo, partial owner match (`notacme/widgets`), repo-name substring (`widgets-internal`), empty URL

`go test ./internal/argocd/`, `go vet`, and `gofmt` all pass locally.

## Scope note re #160

This PR only makes **application/repo matching** work for non-github.com hosts. It does **not** add the GitLab API client, webhook, or runner support that #160 tracks — comment delivery still uses the GitHub API. It's a prerequisite/building block, not a full resolution of #160.

## Notes / alternatives considered

- Kept the change purely additive rather than replacing the github.com-specific candidates, to minimize behavioral risk for existing users.
- Happy to gate this behind an env var / config flag instead if you'd prefer it be opt-in — let me know.

### DIFF
--- a/internal/argocd/helper.go
+++ b/internal/argocd/helper.go
@@ -261,6 +261,25 @@ func gitRepoMatch(repoUrl, repoOwner, repoName string) bool {
 			return true
 		}
 	}
+	// Fallback for non-github.com hosts (GitHub Enterprise, AWS CodeConnections,
+	// mirrors, etc.): match the owner/repo path suffix regardless of host. The
+	// leading separator ('/' for HTTP(S) paths, ':' for scp-style SSH) anchors
+	// the owner boundary so a URL ending in "…/otherowner/repo" is not matched
+	// for owner "owner". Comparison is case-insensitive since git hosting
+	// providers treat owner and repo names case-insensitively.
+	repoUrlLower := strings.ToLower(repoUrl)
+	for _, sep := range []string{"/", ":"} {
+		fallbacks := []string{
+			fmt.Sprintf("%s%s/%s.git", sep, repoOwner, repoName),
+			fmt.Sprintf("%s%s/%s", sep, repoOwner, repoName),
+		}
+		for _, fallback := range fallbacks {
+			if strings.HasSuffix(repoUrlLower, strings.ToLower(fallback)) {
+				log.Debug().Msgf("gitRepoMatch() - non-github host suffix match: %q ends with %q", repoUrl, fallback)
+				return true
+			}
+		}
+	}
 	return false
 }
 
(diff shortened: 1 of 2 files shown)


vrivellino:
@nakul-detroja thank you for your contribution!

I do have one concern: `gitRepoMatch()` is called on every source's `RepoURL` in multi-source apps (`helper.go:103`, `helper.go:201`) regardless of source type. Since the fallback is host-agnostic, a Helm/OCI chart source (e.g. `oci://registry.internal/acme/widgets`) could now coincidentally match the same `owner/repo` suffix as the git source it's paired with, causing the wrong source to get the revision substitution / be included in the diff.

How about we skip the fallback (or skip `gitRepoMatch` generally) when `appSpecSource.Chart != ""`, since that RepoURL isn't a git remote? Could also use a test with a mixed git+chart multi-source app to pin down the intended behavior.

vrivellino:
Also - you can ignore the Docker build failure

vrivellino:
@nakul-detroja: I went ahead and pulled your changes into my own feature branch. There I added the `appSpecSource.Chart != ""` guard as well as an env var to opt out of the matching logic. See #270

vrivellino:
Release via [v2.11.0](https://github.com/vince-riv/argo-diff/releases/tag/2.11.0)

PR #287, opened 2026-08-21
Handle GitHub App installation tokens passed via GITHUB_TOKEN

HelenBaoHB:
## Problem

When `GITHUB_TOKEN` contains a GitHub App installation token (common in CI systems like Jenkins), the startup connectivity check crashes:

```
GET https://api.github.com/user: 403 Resource not accessible by integration
Connectivity check to Github API failed
```

`GET /user` only works with PATs and OAuth tokens. GitHub App installation tokens get 403.

## Fix

When `Users.Get()` returns 403, log a warning and continue instead of crashing. The actual PR diff and commenting endpoints work fine with App tokens — only this health check breaks.

`commentLogin` will be empty, which means argo-diff won't match its own previous comments for updates. It will create new comments instead. This is acceptable — the tool still works.

## Context

argo-diff already supports GitHub Apps via `GITHUB_APP_ID` + `GITHUB_APP_INSTALLATION_ID` + `GITHUB_APP_PRIVATE_KEY`. But many CI systems (Jenkins, etc.) mint an installation token internally and pass it as `GITHUB_TOKEN`. This common pattern hits the bug because the code treats it as a PAT.

### DIFF
--- a/internal/github/comment.go
+++ b/internal/github/comment.go
@@ -155,6 +155,13 @@ func getCommentUser(ctx context.Context) error {
 			log.Info().Msgf("%s received when calling client.Users.Get() via go-github", resp.Status)
 		}
 		if err != nil {
+			// GitHub App installation tokens passed via GITHUB_TOKEN can't
+			// call GET /user (403). This is only a connectivity check —
+			// PR commenting works fine without commentLogin being set.
+			if resp != nil && resp.StatusCode == http.StatusForbidden {
+				log.Warn().Msg("GET /user returned 403 — GITHUB_TOKEN may be a GitHub App installation token. Skipping user lookup.")
+				return nil
+			}
 			log.Error().Err(err).Msg("Unable to determine get my github user")
 			return err
 		}


vrivellino:
@HelenBaoHB: can you tell me a little more about how you're running argo-diff?

When running in server-mode, argo-diff is designed to run with the Github application's private key and it will mint its own installation tokens, as these expire every hour. It would be inappropriate to run argo-diff in server mode with an installation token set `GITHUB_TOKEN`.

Are are you instead invoking it with a fresh installation token on every PR, similar to how it operates in Github actions mode? If so, I think adding an env var to toggle off the connectivity check entirely would be OK.

HelenBaoHB:
> @HelenBaoHB: can you tell me a little more about how you're running argo-diff?
> 
> When running in server-mode, argo-diff is designed to run with the Github application's private key and it will mint its own installation tokens, as these expire every hour. It would be inappropriate to run argo-diff in server mode with an installation token set `GITHUB_TOKEN`.
> 
> Are are you instead invoking it with a fresh installation token on every PR, similar to how it operates in Github actions mode? If so, I think adding an env var to toggle off the connectivity check entirely would be OK.

Hey @vrivellino, we invoke argo-diff as a CLI tool per PR from Jenkins, similar to GitHub Actions mode. Each invocation gets a fresh GitHub App installation token via the Jenkins GitHub App plugin, passed as GITHUB_TOKEN. We don't run it in server mode.

vrivellino:
> > @HelenBaoHB: can you tell me a little more about how you're running argo-diff?
> > When running in server-mode, argo-diff is designed to run with the Github application's private key and it will mint its own installation tokens, as these expire every hour. It would be inappropriate to run argo-diff in server mode with an installation token set `GITHUB_TOKEN`.
> > Are are you instead invoking it with a fresh installation token on every PR, similar to how it operates in Github actions mode? If so, I think adding an env var to toggle off the connectivity check entirely would be OK.
> 
> Hey @vrivellino, we invoke argo-diff as a CLI tool per PR from Jenkins, similar to GitHub Actions mode. Each invocation gets a fresh GitHub App installation token via the Jenkins GitHub App plugin, passed as GITHUB_TOKEN. We don't run it in server mode.

Ok - I'll surface another env var you can use to bypass the connectivity check.

HelenBaoHB:
Thanks for the quick fix in #293 @vrivellino ! Any idea when the next release will be cut? We'd like to start using ARGO_DIFF_BYPASS_CONNECTIVITY_CHECKS in our Jenkins pipeline.

vrivellino:
> Thanks for the quick fix in #293 @vrivellino ! Any idea when the next release will be cut? We'd like to start using ARGO_DIFF_BYPASS_CONNECTIVITY_CHECKS in our Jenkins pipeline.

I just cut [2.13.2](https://github.com/vince-riv/argo-diff/releases/tag/2.13.2)!

Please note from the readme:

`ARGO_DIFF_BYPASS_CONNECTIVITY_CHECKS`:
> Comma-separated list of startup connectivity checks to skip: `github`, `argocd`, `true`/`all` for both, or `false`/`none` (default, skips nothing). Useful when `GITHUB_TOKEN` holds a GitHub App installation token, which can't call `GET /user` (a per-PR CI invocation minting its own token, eg: from Jenkins, is the common case — GitHub Actions mode doesn't need this, it already skips the check); for that case, bypass `github` specifically rather than `true`/`all`. Bypassing `github` also makes argo-diff match its own prior PR comments by marker only, the same as it does under GitHub Actions — so two instances sharing an empty `ARGO_DIFF_CONTEXT_STR` on one PR will overwrite each other's comments. Bypassing `argocd` skips reachability **and** the minimum ArgoCD version check (client and server must be `2.12.0`+), so it also silently drops that guard.

PR #289, opened 2026-08-22
feat: diff ArgoCD applications concurrently

### LINKED ISSUE #273: Diff ArgoCD applications in parallel with a bounded worker pool
### Problem

`GetApplicationChanges()` diffs matching applications strictly sequentially, and each
`argocd app diff` is almost entirely time spent waiting on the argocd server. Wall-clock time
therefore grows linearly with the number of matching applications, which is exactly the case a
shared chart in a monorepo produces.

Measured in a real GitHub Actions run (a chart change in a monorepo, referenced by relative
path from many applications):

- 49 matching multi-source applications, ~1.5s per diff
- ~147s of diffing, against a 3 minute budget for the whole event

#272 removed a duplicate-diff bug that was doubling this, and made the timeout configurable, so
that run now fits. But the headroom is proportional to app count: another ~50 applications
sharing the chart puts it back over the default, and raising the timeout just trades a failed
run for a slow one.

### Proposal

Diff applications concurrently with a bounded worker pool (a `golang.org/x/sync/errgroup` with
`SetLimit`, or a plain semaphore channel), defaulting to something conservative like 4-8 and
configurable via env var. Applied to both loops in `GetApplicationChanges()` — the single-source
loop and the multi-source loop — this should cut diff wall-clock roughly linearly in the pool
size.

### Things to work out

- **Load on the argocd server.** N concurrent `argocd app diff` calls means N concurrent manifest
  generations on the repo-server. The default needs to be safe for a modest ArgoCD install, and
  the knob wants documenting alongside `ARGOCD_OPTS`.
- **Ordering.** Results currently accumulate in `filterApplications()` order, which determines
  the order of sections in the PR comment. Collect into an indexed slice rather than appending
  from workers, so comments stay stable between runs.
- **The nested-application path.** The single-source loop calls `argoAppsWithChanges()` and then
  diffs any nested multi-source applications it discovers, appending to the shared
  `multiSrcAppNamesDiffed` slice that the second loop reads. That sharing needs to be made safe
  (or restructured into a discovery phase followed by a diff phase) before the loops can run
  concurrently.
- **Interaction with the timeout reserve.** The `ctx.Err()` bail-out and `notDiffed` reporting
  added in #272 assume one diff in flight at a time; with a pool, in-flight workers need to be
  drained and attributed correctly.

### Alternative considered

Raising `ARGO_DIFF_TIMEOUT` is the workaround available today, and is fine for moderate app
counts. It doesn't help the PR feedback loop, which is the point of the tool — a 5+ minute wait
for a diff comment is a poor experience regardless of whether it eventually succeeds.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

brushmtn:
## Summary
- Parallelize `GetApplicationChanges()` in `internal/argocd/helper.go` using a bounded worker pool, so ArgoCD application diffs run concurrently instead of sequentially.
- Add `ARGO_DIFF_MAX_WORKERS` (default 4, capped at 32) to control the pool size, wired through `README.md`, `action.yml`, and the Helm chart (`values.yaml`, `configmap.yaml`, chart tests, chart `README.md`).
- Preserve existing behavior: deterministic merged output order, partial-result/`notDiffed` handling on context timeout, and a pre-existing wave-2 `WarnStr` inconsistency (documented in `internal/argocd/context.md` as a known follow-up, not fixed here).
- The diff runs in three sequential waves (top-level apps, then nested app-of-apps, then remaining multi-source apps), each bounded by the same worker limit, so at most `ARGO_DIFF_MAX_WORKERS` `argocd` CLI calls are in flight system-wide.

Closes #273.

## Test plan
- [x] `go build -v ./...`
- [x] `go test ./...` (including `-race` on the new/changed `internal/argocd` tests)
- [x] `golangci-lint run` (no new issues)
- [x] `helm lint` / `helm unittest .` in `charts/argo-diff`
- [x] `helm-docs --chart-search-root=charts` regenerated `charts/argo-diff/README.md`

<details>
<summary>Original design plan</summary>

```markdown
# Parallel ArgoCD Application Diffing (issue #273)

## Context

`GetApplicationChanges()` (`internal/argocd/helper.go:119-258`) diffs every matching ArgoCD
application one at a time. In a real-world monorepo where 49 applications share a chart, this took
~147s against the default 3-minute (`ARGO_DIFF_TIMEOUT`) budget — leaving little headroom before
runs start failing as more applications are added. Issue #273 asks for a bounded worker pool so
diffing scales with wall-clock time close to `apps / workers` instead of `apps`, with a
conservative, configurable concurrency limit so the ArgoCD repo-server isn't overloaded by a burst
of concurrent manifest generations.

Decisions locked in with the user:
- Hand-rolled semaphore + `sync.WaitGroup` pool (no new dependency; avoids `errgroup`'s
  cancel-on-first-error semantics, which don't fit since per-app failures here are expected and
  must not abort sibling in-flight diffs).
- Nested app-of-apps children are parallelized too, not just the top-level app loop.
- Default worker count: `4`, via a new `ARGO_DIFF_MAX_WORKERS` env var.
- Ship first-class Helm chart support (`values.yaml` + `configmap.yaml` + chart unittest), not just
  README/`action.yml`.

## Design: three sequential bounded "waves"

Naively giving each parent-app goroutine its own nested-app sub-pool risks deadlock (a goroutine
holding one of N semaphore slots blocking forever trying to acquire another when N is small) and
makes the concurrency bound unpredictable (total in-flight work could multiply across levels).
Instead, restructure `GetApplicationChanges()` into three waves, each bounded by the same
`runWithLimit(n, limit, work)` helper, each running to completion (`wg.Wait()`) before the next
starts — this keeps at most `limit` diff operations in flight at any instant, system-wide, and
avoids recursive pool usage entirely.

**Wave 1 — top-level single-source apps** (replaces `helper.go:147-207`'s outer loop body):
For each `app`, a worker: checks `ctx.Err()` at dispatch time, calls `getApplicationChanges()`
(today's own-app diff), and if it has changes, calls `argoAppsWithChanges()` to enumerate nested
Application manifests. For each nested `subApp` found in `appLookup`, it **unconditionally** records
the name (preserving today's `helper.go:186` behavior of reserving the name before checking
`ctx.Err()`) and, if `ctx` is still alive, queues a `nestedJob{appCur, appNew}` for wave 2 instead of
diffing it inline. Each worker returns a private result struct (own diff outcome, `notDiffed`
entries, reserved sub-app names, queued nested jobs) — no shared-state writes inside the worker.
After `runWithLimit` drains, the orchestrator merges all workers' results **in original `apps`
order** into `appResList`, `notDiffed`, and `multiSrcAppNamesDiffed`, and flattens all queued
`nestedJob`s (in that same parent order) into one list for wave 2.

**Wave 2 — nested app diffing, flat across all parents** (replaces the inline `subApp` loop at
`helper.go:184-204`): bounded pool over the flattened `nestedJob` list. Each worker re-checks
`ctx.Err()` (time may have passed since wave 1 queued it) and calls `getMultiSrcAppChanges()`.
Preserve today's exact behavior bug-for-bug: on a non-timeout error, `WarnStr` is set on the result
but — matching current `helper.go:191-200` — is **not** appended to `appResList` (only the
`len(...) > 0` success branch appends). This looks like a pre-existing inconsistency (the top-level
loop *does* append its `WarnStr` case), but fixing it is out of scope for this issue; flag it to the
user as a possible follow-up, don't fix it silently here. Merge results into `appResList`/`notDiffed`
in flattened-job order after the pool drains.

**Wave 3 — multi-source apps** (replaces `helper.go:223-252`): unchanged logic, now pool-bounded.
The `slices.Contains(multiSrcAppNamesDiffed, ...)` skip check (today `helper.go:224`) runs in the
orchestrator *before* dispatch (a sequential filter over the re-filtered `apps` list, since
`multiSrcAppNamesDiffed` is fully populated by the end of wave 1 — the same natural barrier that
exists today, just now enforced by `runWithLimit`'s `wg.Wait()` instead of "the loop finished
iterating"). Remaining apps go through the bounded pool; results merge in original order.

**Ordering impact**: `appResList` changes from "parent, its nested apps, next parent, ..." to "all
parents (wave 1, original order), all nested apps (wave 2, flattened parent order), multi-source
apps (wave 3)". Verified via `Explore` agent: nothing in `internal/process_event/code_change.go`,
`code_change_test.go`, or `internal/github`'s markdown builder (`CommentMarkdown.String()`) depends
on parent/nested adjacency — each app becomes an independent `<details>` block, and the only
existing test touching this path (`TestGetApplicationChangesOutOfTimeEnumeratingNestedApps`) checks
counts, not order. Safe to proceed.

## Files to change

### `internal/argocd/concurrency.go` (new)

- `runWithLimit(n, limit int, work func(i int))` — semaphore-channel + `sync.WaitGroup` pool, `limit`
  clamped to `>= 1`.
- `maxWorkers() int` — reads `ARGO_DIFF_MAX_WORKERS`, modeled on `processTimeout()`
  (`internal/process_event/code_change.go:24-40`): plain function (not `init()`-cached, so
  `t.Setenv` works in tests), positive-int validation, default `4`, capped at `32` (guards against
  an accidental huge value hammering the repo-server), warns and falls back on invalid input.

### `internal/argocd/helper.go`

- Add `nestedJob`, and small per-wave result structs, near `ApplicationResourcesWithChanges`.
- Extract wave-1 and wave-2 worker bodies into standalone functions taking only read-only inputs
  (`ctx`, `app`, `appLookup`, `eventInfo`) and returning a private result struct — no closures over
  shared mutable state.
- Rewrite `GetApplicationChanges()` to call `maxWorkers()` once, then run the three waves via
  `runWithLimit`, merging each wave's indexed results sequentially before the next wave starts.
- Keep `getApplicationChanges()` / `getMultiSrcAppChanges()` (`helper.go:70-110`) as-is — they're
  already pure enough to call from workers unchanged.

### `internal/argocd/helper_test.go`

- `TestGetApplicationChangesOutOfTime` and `TestGetApplicationChangesOutOfTimeEnumeratingNestedApps`:
  add `t.Setenv("ARGO_DIFF_MAX_WORKERS", "1")` so wave 1 runs one app at a time, preserving the
  strict program order the second test's mock relies on (it calls the test's own `cancel()` from
  inside the diff callback to simulate the deadline landing mid-run).
- New test: pool bounds concurrency — mock `execArgoCdCli` tracks current/max in-flight via
  `atomic.Int64` plus a small `time.Sleep`, against a fixture with more matching apps than the
  configured limit; assert observed max concurrency equals the configured `ARGO_DIFF_MAX_WORKERS`
  (and is `> 1`, proving it's actually parallel).
- New test: order stability — run with `ARGO_DIFF_MAX_WORKERS=4` against a multi-app fixture with
  deterministic per-app mock outputs; assert `appResList` order matches a `MAX_WORKERS=1` run.
- Audit other `execArgoCdCli =` mock assignments in this file for unsynchronized shared state
  (bare `int`/slice closures) that a concurrent path would now race on.

### Documentation / deployment surfaces

- **`README.md`** — new row in the env var table, between `ARGO_DIFF_DISABLE_NON_GITHUB_REPO_MATCH`
  (line 230) and `ARGO_DIFF_TIMEOUT` (line 231): `ARGO_DIFF_MAX_WORKERS`, action input
  `max_workers`, default `4`, description covering the concurrency bound and noting it works
  alongside `ARGOCD_OPTS` (e.g. raising the `argocd` CLI's own `--timeout`) if increased, since more
  concurrent diffs means more concurrent repo-server load.
- **`action.yml`** — add `max_workers` input (alphabetically after `log_level`, before
  `repo_default_ref`, ~line 67-70) and `ARGO_DIFF_MAX_WORKERS: ${{ inputs.max_workers }}` to the
  `runs.env` block (~line 83, alphabetically between `ARGO_DIFF_CONTEXT_STR` and `ARGO_DIFF_TIMEOUT`).
- **`charts/argo-diff/values.yaml`** — add `config.maxWorkers: ""` with a `# config.maxWorkers --`
  helm-docs comment, alongside `commentLineMaxChars` etc.
- **`charts/argo-diff/templates/configmap.yaml`** — add `.maxWorkers` to the `if or` condition list
  and a `{{- with .maxWorkers }} ARGO_DIFF_MAX_WORKERS: {{ . | quote }} {{- end }}` block, matching
  the existing `commentLineMaxChars`/`ARGOCD_OPTS`-style entries.
- **`charts/argo-diff/tests/configmap_test.yaml`** — add a `config maxWorkers` case asserting
  `data.ARGO_DIFF_MAX_WORKERS` matches the set value, matching the existing test shape.
- Regenerate `charts/argo-diff/README.md` via `helm-docs` (per `charts/context.md`) so the
  chart's own README table picks up the new values key.

## Verification

- `go build -v ./...` and `go test ./...` (in particular `go test ./internal/argocd/...` with
  `-race` to catch any missed synchronization in the new pool/result-merge code).
- `golangci-lint run`.
- `helm unittest charts/argo-diff` (or however the repo's `helm.yml` workflow invokes chart tests)
  for the new `configmap_test.yaml` case.
- Manually run `go run cmd/main.go -f event.json` against a local/dev ArgoCD with several matching
  applications, comparing PR-comment output and timing with `ARGO_DIFF_MAX_WORKERS=1` (should match
  today's sequential output/order) vs. a higher value (should be faster, same app coverage).
```

</details>


vrivellino:
# Code review: parallel app diffing

**One blocker: the PR activates a latent argv-aliasing race that makes argo-diff diff the wrong applications, silently.** The rest of the change is sound.

---

## 🔴 Blocker — `execArgoCdCli` shares one backing array across workers

`internal/argocd/argocd_client.go:88`:

```go
argv := append(commonCliArgv, args...)
```

`commonCliArgv` is built by repeated `append` in `init()`. Some configurations leave it with **spare capacity**. When that happens, `append` writes the per-call args into the *shared* backing array. Sequential code never noticed. The worker pool does.

I confirmed it against this branch. With `ARGOCD_SERVER_INSECURE`, `ARGOCD_SERVER_PLAINTEXT`, `ARGOCD_GRPC_WEB` and `ARGOCD_GRPC_WEB_ROOT_PATH` all set, `commonCliArgv` is `len=9 cap=16`:

```
WARNING: DATA RACE
ARGV CORRUPTED: wanted [app diff pool-app-3 ...] got [app diff pool-app-4 ...]
ARGV CORRUPTED: wanted [app diff pool-app-2 ...] got [app diff pool-app-4 ...]
ARGV CORRUPTED: wanted [app diff pool-app-0 ...] got [app diff pool-app-4 ...]
   (8 of 8 workers diffed pool-app-4)
```

**Impact:** every worker runs the same `argocd app diff`. Argo-diff posts one app's diff under many app names. No error, no warning — just a wrong PR comment.

**Fix** (one line):

```go
argv := slices.Concat(commonCliArgv, args)
```

**Please also add a test that exercises the real argv construction.** Every current test replaces `execArgoCdCli` wholesale, so no test can ever catch this class of bug.

Two things make this worse than the narrow trigger suggests:

- Which configs are affected depends entirely on Go's slice growth. Add one flag to `init()` and a different set of users breaks.
- Server mode already runs each event in its own goroutine, so concurrent events can already hit this. This PR makes it reachable from a **single** event.

---

## 🟡 Medium — app-of-apps grouping is lost in the PR comment

Order changes from *"parent, its children, next parent"* to *"all parents, then all children, then multi-source"*. I verified nothing downstream breaks: `code_change.go:167` and `markdown.go:141` just iterate, and each app is its own `<details>` block.

But for app-of-apps users the comment is now harder to read — a parent and its children sit far apart.

Cheap fix: add `parentIdx` to `nestedJob`, then merge wave-2 results next to their parent instead of appending them as one block.

---

## 🟢 Verified good

- `go test -race -count=3 ./...` — clean across all packages.
- `go vet ./...` — clean.
- The index-addressed result merge is the right pattern. Order determinism is genuinely proven by `TestGetApplicationChangesOrderStable`.
- No shared mutable state in the workers. `GetSource()`/`GetSources()` are read-only; `appLookup` is read-only; `os.Getenv` is safe.
- Timeout and `notDiffed` semantics match the sequential version. Wave-1 name reservation before the ctx check is preserved.
- Env var wired through all four surfaces per `AGENTS.md`: README, `action.yml`, chart values/template/unittest. `docs/k8s/` needs nothing — it carries no optional env vars.
- `action.yml` default `''` correctly falls through to the Go default of 4.
- The wave-2 `WarnStr` inconsistency is documented, not silently changed. Good call.

---

## 🔵 Nits

- `TestGetApplicationChangesConcurrencyBound` asserts `maxSeen == workers` exactly. That can under-report on a loaded CI runner. `>= 2 && <= workers` proves both parallelism and the bound without the flake risk. It passed 10 runs at `-cpu=1` here, so risk is low.
- Four concurrent diffs hold four full manifest sets in memory. The chart ships `resources: {}`, so no default OOM cliff — just worth knowing if users set limits and raise `maxWorkers`.

---

**Recommendation: request changes.** Fix the `append` aliasing, then this is good to merge.

---

brushmtn:
Addressed the review in 36f3e95:

**🔴 Blocker (argv aliasing race)** — fixed. `execArgoCdCli` now builds `argv` with `slices.Concat(commonCliArgv, args)` instead of `append(commonCliArgv, args...)`, so it always allocates a fresh backing array instead of writing into `commonCliArgv`'s spare capacity. Added `TestExecArgoCdCliConcurrentArgvIsolation`, which reproduces a `commonCliArgv` with spare capacity and asserts many concurrent `execArgoCdCli` calls each see exactly their own args (via a fake `argocd` script driven by `ARGOCD_CLI_CMD_NAME`). Confirmed it fails under `-race` against the old `append`-based code and passes with the fix.

**🟡 Medium (app-of-apps grouping)** — fixed. `nestedJob` now carries `parentIdx`, and `GetApplicationChanges()` merges wave-1 and wave-2 results back together so each parent's diff is immediately followed by its own nested apps' diffs, restoring the pre-parallelization "parent, its nested apps, next parent, ..." order. Added `TestGetApplicationChangesNestedAppsGroupedWithParent` (multiple parents with nested apps, diffed concurrently with artificial jitter, still comes back grouped); confirmed it fails against the old flat-merge behavior.

**🔵 Nits** — took the concurrency-bound suggestion: `TestGetApplicationChangesConcurrencyBound` now asserts `>1 && <=workers` instead of exact equality, to avoid under-reporting on a loaded CI runner. Left the memory-per-worker note as-is per your comment (no default `resources` limit in the chart, so no action needed there).

`go build`, `go vet`, `gofmt`, and `go test -race -count=3 ./...` all pass. `context.md` updated to describe the argv-aliasing hazard and the parent/child merge.

vrivellino:
Both fixes verified in `36f3e95`. **Approving.**

**🔴 Blocker — fixed.** `slices.Concat` is the right call. I reverted it to the old `append` and the new test caught it:

```
WARNING: DATA RACE
--- FAIL: TestExecArgoCdCliConcurrentArgvIsolation
    worker 14: argv corrupted: wanted [app diff pool-app-14] got [app diff pool-app-6]
```

Nice touch setting `cap=16, len=9` explicitly rather than relying on Go's growth algorithm — the test stays valid if a flag is ever added to `init()`. And it drives a real subprocess, so it covers the argv path nothing else did.

**🟡 Medium — fixed.** Checked the invariant the merge relies on: `nestedJobs` accumulates in `wave1Results` index order, so it is ascending and contiguous by `parentIdx`, and the single-pointer walk consumes every job. Restoring the flat merge fails the new test:

```
got  [pool-app-0 pool-app-1 pool-app-2 pool-app-0-child pool-app-2-child]
want [pool-app-0 pool-app-0-child pool-app-1 pool-app-2 pool-app-2-child]
```

**🔵 Nit — fixed.**

My verification: `go test -race -count=3 ./...` passes; new tests survive 15 runs at `-cpu=1` with no flakes; `go vet` and `gofmt -l` clean. Documenting the `append` hazard as a standing rule in `context.md` is the part that stops this recurring — good call.

Last nit, non-blocking: `notDiffed` still merges as "all wave-1 names, then all wave-2 names", so it no longer matches `appResList`'s parent-grouped order. Cosmetic; fine to leave.

---

brushmtn:
Took the last nit too, in 6f6eea8: `notDiffed` now merges through the same `parentIdx`-ordered walk as `appResList`, so a nested app's skip sits next to its parent's entry instead of in a separate "all wave-1 skips, then all wave-2 skips" block. Added `TestGetApplicationChangesNotDiffedGroupedWithParent`, which reproduces the one scenario where the old flat merge and the new grouped merge actually diverge (a later top-level app skipped outright, positioned after an earlier parent whose nested child was also skipped) — confirmed it fails against the old flat merge and passes with the fix.

`go build`, `go vet`, `gofmt`, and `go test -race -count=3 ./...` all still pass.

vrivellino (approved):

PR #293, opened 2026-08-24
feat: add ARGO_DIFF_BYPASS_CONNECTIVITY_CHECKS env var

vrivellino:
## Problem

PR #287 reports argo-diff crashing on startup when `GITHUB_TOKEN` holds a GitHub App installation token (common when a CI system like Jenkins mints one per-PR, outside GitHub Actions). `GET /user` 403s for installation tokens, and the startup connectivity check turns that into a fatal exit.

PR #287's fix swallows the 403 inside `getCommentUser()`, but that leaves `commentLogin` empty, breaking comment matching — argo-diff would post a new comment every run instead of updating its own.

## Fix

New env var: `ARGO_DIFF_BYPASS_CONNECTIVITY_CHECKS` — comma-separated, values `github`, `argocd`, or `true`/`all` for both. Unrecognized values log a warning and are ignored.

Bypassing `github` skips the 403-prone call everywhere it happens, not just at startup — also in `getExistingComments()`, which runs on every event — and falls back to matching argo-diff's own prior comments by marker only, exactly like GitHub Actions mode already does. This avoids the duplicate-comment problem the alternative fix would cause.

## Changes

- `internal/config/` (new package): env var parsing, with tests.
- `cmd/main.go`: skips `argocd.ConnectivityCheck()` when bypassed for `argocd`.
- `internal/github/comment.go`: skips the GitHub connectivity check and comment-author match when bypassed for `github`.
- `internal/server/run_once.go`: new var added to debug env logging.
- README, `AGENTS.md`, and `internal/github/context.md` updated.
- Regression test `TestCommentExistingDifferentUserBypass` reproduces the installation-token case and confirms the existing comment is found and updated.

Closes the underlying issue in #287.


internal/github/comment.go
```diff
@@ -276,7 +289,7 @@ func getExistingComments(ctx context.Context, owner, repo string, prNum int) ([]
 		log.Debug().Msgf("Checking %d comments in %s/%s#%d", len(comments), owner, repo, prNum)
 		for _, c := range comments {
 			if strings.Contains(*c.Body, commentIdentifier) {
-				if isGithubAction || *c.User.Login == commentLogin {
+				if isGithubAction || bypassGithubCheck() || *c.User.Login == commentLogin {
```
brushmtn: Nit (efficiency): `bypassGithubCheck()` runs once per comment per page here, and each call does an `os.Getenv`, a string split, and a fresh map allocation in `parseBypassList`. Line 270 already computed the same value earlier in this function, so a local would cover both sites:

```go
bypass := bypassGithubCheck()
if !isGithubAction && !bypass {
	err := getCommentUser(ctx)
	...
}
...
if isGithubAction || bypass || *c.User.Login == commentLogin {
```

The env var can't change mid-call, so hoisting it doesn't alter behavior.

README.md
```diff
@@ -225,6 +225,7 @@ the accepted environment variables and their respective GitHub Actions inputs.
 | ARGOCD_SERVER_INSECURE           | argocd_server_insecure      | no               | `false`  | Set `--insecure` flag for argocd cli (`true`/`false`). |
 | ARGOCD_SERVER_PLAINTEXT          | argocd_server_plaintext     | no               | `false`  | Set `--plaintext` flag for argocd cli (`true`/`false`). |
 | ARGOCD_UI_BASE_URL               | argocd_ui_base_url          | no               |          | Base URL of ArgoCD UI (usually the server name prefixed with `https://`). |
+| ARGO_DIFF_BYPASS_CONNECTIVITY_CHECKS | N/A                      | no               |          | Comma-separated list of startup connectivity checks to skip: `github`, `argocd`, or `true`/`all` for both. Useful when `GITHUB_TOKEN` holds a GitHub App installation token, which can't call `GET /user` (a per-PR CI invocation minting its own token, eg: from Jenkins, is the common case — GitHub Actions mode doesn't need this, it already skips the check). Bypassing `github` also makes argo-diff match its own prior PR comments by marker only, the same as it does under GitHub Actions — so two instances sharing an empty `ARGO_DIFF_CONTEXT_STR` on one PR will overwrite each other's comments. |
```
brushmtn: `argocd.ConnectivityCheck()` does two jobs, and this row only describes one of them. Besides reaching the server, it enforces `minVersion` on both the client and the server (`internal/argocd/helper.go:48-59`) — so bypassing `argocd` silently drops the minimum-version guard too, not just a reachability probe.

That matters because `true` reads like a plain boolean. An operator who hits the `GET /user` 403 and reaches for `ARGO_DIFF_BYPASS_CONNECTIVITY_CHECKS=true` gets the GitHub fix they wanted plus a disabled argocd version check they didn't ask for, with no signal beyond the startup warning.

Suggest calling that out here and steering the documented remedy for the installation-token case to `github` specifically rather than `true`/`all`.

internal/config/connectivity.go
```diff
@@ -0,0 +1,85 @@
...
+		tok = strings.ToLower(strings.TrimSpace(tok))
+		if tok == "" {
+			continue
+		}
+		switch tok {
+		case "true", "all":
+			for _, c := range allComponents {
+				bypassed[c] = true
+			}
+		case "false", "none":
```
brushmtn, line 43: `false` and `none` are accepted as recognized no-ops here, but the README table (and so the user-facing docs) lists only `github`, `argocd`, `true`, and `all`. `internal/config/context.md` documents them, which agents will find, but an operator reading the README who sets `ARGO_DIFF_BYPASS_CONNECTIVITY_CHECKS=false` has no way to confirm that's a supported value rather than an ignored typo.

Worth adding `false`/`none` to the README row's list of accepted values.

brushmtn (commented):
Three small follow-ups from a review pass — one efficiency nit and two doc gaps. Nothing blocking: the fix itself looks correct, all three `getCommentUser`/`commentLogin` call sites are covered, `go build`, `go test ./...`, `gofmt` and `golangci-lint` are clean on this branch, and `TestCommentExistingDifferentUserBypass` is a genuine regression test (the PR-2 fixture comment is authored by `baduser` while `/user` returns `testuser`, so it fails without the change).

internal/github/comment.go
brushmtn: Confirmed in c1c330c — `bypass` is computed once and used at both sites, and behavior is unchanged. Resolving.

README.md
brushmtn: Confirmed in c1c330c. The row now states that bypassing `argocd` drops the minimum-version guard as well as reachability, and steers the installation-token case to `github` specifically. I checked `internal/argocd/helper.go:20` — `minVersion = "2.12.0"`, so the version cited is accurate. Resolving.

internal/config/connectivity.go
brushmtn: Confirmed in c1c330c — the README row now lists `false`/`none` alongside the other accepted values. Resolving.

PR #295, opened 2026-08-24
fix(github): derive App comment login from slug, not display name

### LINKED ISSUE #290: GitHub App comment author is derived from App name instead of slug
## Summary

When argo-diff is configured as a GitHub App, `getCommentUser()` derives the bot login from the
App's **display name** instead of its **slug**. If the App name is not already slug-shaped, the
computed login never matches the real comment author, and argo-diff stops recognizing its own
previous comments.

## Where

`internal/github/comment.go:150`

```go
commentLogin = *app.Name + "[bot]"
```

## Why it's wrong

GitHub builds a bot's login from the App **slug** — the URL-friendly form of the name (lowercased,
spaces replaced with hyphens) — not from the display name.

| App display name | Real comment author | Value computed today |
| ---------------- | ------------------- | -------------------- |
| `argo-diff`      | `argo-diff[bot]`    | `argo-diff[bot]` ✅ |
| `Argo Diff`      | `argo-diff[bot]`    | `Argo Diff[bot]` ❌ |
| `ArgoDiff Prod`  | `argodiff-prod[bot]`| `ArgoDiff Prod[bot]` ❌ |

`go-github` already exposes the correct field — `App.Slug` alongside `App.Name`
(`github/apps.go:22`).

## Impact

`getExistingComments()` filters candidate comments with:

```go
if isGithubAction || *c.User.Login == commentLogin {
```

(`internal/github/comment.go:279`)

When `commentLogin` is wrong, that comparison never matches. The consequences:

- argo-diff never finds its previous comment on a PR.
- Every event creates a **new** full-size diff comment instead of editing the existing one.
- The `[Outdated argo-diff content]` cleanup path never runs, so stale comments accumulate.

The symptom is duplicate comments on every push to a PR.

This is latent rather than universal because `README.md` recommends naming the App `argo-diff`,
which is already slug-shaped. Any other name silently breaks comment reuse. Nothing in the setup
docs warns that the name matters.

## Suggested fix

Use the slug, and keep the name only as a fallback:

```go
if app == nil || (app.Slug == nil && app.Name == nil) {
	log.Error().Msg("Empty app returned - not sure how I got here")
	return fmt.Errorf("empty app info")
}
appLogin := app.GetSlug()
if appLogin == "" {
	log.Warn().Msg("Github App slug is empty - falling back to app name")
	appLogin = app.GetName()
}
mux.Lock()
commentLogin = appLogin + "[bot]"
mux.Unlock()
```

Note the nil check at `internal/github/comment.go:145` also needs updating, since it currently
guards only `app.Name`.

## Test coverage gap

There is no test for the GitHub App identity path. `internal/github/github_testdata/` has
`payload-user.json` for the PAT path but no `/app` fixture, and no test exercises
`commentClientIsApp = true`. A fixture where slug and name differ would pin this behavior down.

brushmtn:
Fixes #290. `getCommentUser()` built the GitHub App bot login from `App.Name` (display name) instead of `App.Slug`, so any App whose display name isn't already slug-shaped (e.g. `Argo Diff` instead of `argo-diff`) never matched its own comment author. This made argo-diff post a new full-size comment on every PR event instead of editing its existing one, and the "Outdated argo-diff content" cleanup never ran.

This PR switches `commentLogin` to use `App.GetSlug()`, falling back to `App.GetName()` if the slug is empty, adds a test fixture/test for the GitHub App identity path (previously untested), and updates the README to note that any App display name now works correctly.

<details>
<summary>Plan</summary>

1. Fix `getCommentUser()` in `internal/github/comment.go` to use `app.GetSlug()` with fallback to `app.GetName()`, and update the nil-check to guard on both fields being nil.

2. Add a `/app` fixture to `internal/github/github_testdata/` where slug and name differ, and add a test exercising `commentClientIsApp = true` to pin down the correct login derivation.

3. Update README.md to remove/adjust the guidance that the GitHub App must be named `argo-diff`, since the slug is now used regardless of display name.

4. Run `go build`, `go fmt`, `golangci-lint run`, and `go test ./...` to verify.

</details>

<details>

All review feedback from @vrivellino addressed, replies posted on all 4 inline threads.

Latest fix (commit `5e86d32`): the `TestGetCommentUserApp` httptest handler used `t.Fatalf` on a fixture-read failure, but `Fatalf` calls `runtime.Goexit`, which must run on the test's own goroutine, not the HTTP handler goroutine `httptest.NewServer` spawns it on. Calling it there would kill the handler mid-response instead of failing the test. Switched to `t.Errorf` plus an explicit 500 response and early return, matching the existing pattern a few lines above in the same handler. `go vet`'s `testinggoroutine` analyzer doesn't catch this case (it doesn't follow goroutines started via `httptest.NewServer`), so this was only caught by manual review.

Earlier fixes (commit `361a4a1`): a real bug where a non-nil-but-empty App slug bypassed the nil guard and could cache `commentLogin` as `"[bot]"` forever (reproducing the #290 symptom) — fixed by validating the derived login value and erroring when both slug and name are empty. Also fixed `TestGetCommentUserApp` depending on test execution order for its `commentClient` setup, and added two more table-driven cases (empty-slug fallback, both-empty error) with new fixtures to close the coverage gap on those branches.

Verified: `go build -v ./...`, `go vet ./...`, `go fmt ./...`, and `go test ./...` all pass. `golangci-lint run ./internal/github/...` shows only pre-existing issues untouched by this PR.

Nothing outstanding. Ready for re-review.

</details>


internal/github/comment.go
```diff
@@ -155,12 +155,17 @@ func getCommentUser(ctx context.Context) error {
...
 			return fmt.Errorf("empty app info")
 		}
+		appLogin := app.GetSlug()
+		if appLogin == "" {
+			log.Warn().Msg("Github App slug is empty - falling back to app name")
+			appLogin = app.GetName()
+		}
 		mux.Lock()
-		commentLogin = *app.Name + "[bot]"
+		commentLogin = appLogin + "[bot]"
```
vrivellino: **Finding 1 — an empty-string slug still produces a broken login.**

The guard checks the *pointers*, but the fallback checks the *value*. An `App` with `Slug` set to `""` (non-nil, empty) passes the guard on line 158, then `GetSlug()` returns `""`, then `GetName()` also returns `""`. `commentLogin` becomes the literal `"[bot]"`.

I confirmed this by running the branch logic directly:

```
commentLogin would be "[bot]"
```

That value is non-empty, so `getCommentUser()` caches it and never retries. Every author match at line 293 then fails — which is exactly the #290 symptom this PR sets out to fix.

Suggested fix: validate the *derived* value rather than the pointers.

```go
appLogin := app.GetSlug()
if appLogin == "" {
    log.Warn().Msg("Github App slug is empty - falling back to app name")
    appLogin = app.GetName()
}
if appLogin == "" {
    log.Error().Msg("Github App has neither slug nor name")
    return fmt.Errorf("empty app info")
}
```

That also makes the widened `app.Slug == nil && app.Name == nil` check redundant — a plain `app == nil` is enough.

internal/github/comment_test.go
```diff
@@ -386,3 +390,42 @@ func TestCommentNotHead(t *testing.T) {
...
+		mux.Lock()
+		commentLogin = origCommentLogin
+		mux.Unlock()
+	}()
+
+	var err error
+	appsClient, err = github.NewClient(github.WithURLs(&baseURL, &baseURL))
+	if err != nil {
+		t.Fatalf("Failed to create github client: %s", err)
+	}
```
vrivellino: **Finding 2 — this test fails when run on its own.**

The test sets `appsClient` but never sets `commentClient`. It only passes today because earlier tests in this file assign that package global first. Run in isolation it fails:

```
$ go test -run TestGetCommentUserApp ./internal/github/...
--- FAIL: TestGetCommentUserApp (0.00s)
    comment_test.go:424: getCommentUser() failed: no github commenter client
```

`getCommentUser()` returns early at the `commentClient == nil` check (comment.go:137) before it ever reaches the App path.

Two reasons this matters: `AGENTS.md` documents `go test -run TestName ./...` as a normal workflow, and the dependency breaks silently if anyone reorders these tests or adds one after this.

Fix: point `commentClient` at the test server inside the test, the way every other test in this file does.

```go
commentClient, err = github.NewClient(github.WithAuthToken("test1234"), github.WithURLs(&baseURL, &baseURL))
if err != nil {
    t.Fatalf("Failed to create github client: %s", err)
}
```

internal/github/comment_test.go
```diff
@@ -386,3 +390,42 @@ func TestCommentNotHead(t *testing.T) {
...
+	mux.Unlock()
+
+	if err := getCommentUser(context.Background()); err != nil {
+		t.Fatalf("getCommentUser() failed: %s", err)
+	}
+
+	want := "argodiff-prod[bot]"
+	if commentLogin != want {
+		t.Errorf("commentLogin = %q, want %q (must come from App slug, not display name %q)", commentLogin, want, "ArgoDiff Prod")
+	}
```
vrivellino: **Finding 3 — the new fallback branch has zero test coverage.**

This PR adds a branch and a test, but the test only exercises the happy path. Coverage confirms the `appLogin = app.GetName()` fallback never runs:

```
$ go test -coverprofile=c.out ./internal/github/
comment.go:163.21,166.4 2 0     <- hit count 0
```

That uncovered branch is precisely where Finding 1 lives, which is why the bug slipped through.

Please add a second case with an empty slug — a fixture where `slug` is `""` and `name` is set — asserting the login falls back to the name. Once Finding 1 is fixed, a third case (both empty) should assert `getCommentUser()` returns an error rather than caching `"[bot]"`.

vrivellino (commented):
Core fix is correct — `App.GetSlug()` is the right source, since GitHub builds a bot login as `<slug>[bot]`. Verified `go build`, `gofmt`, and `go test ./...` all pass on this branch. Three findings inline before merge.

internal/github/comment.go
brushmtn: Fixed in 361a4a1 — validated the derived `appLogin` value instead of the source pointers, and added the both-empty error path so it no longer caches a bare `"[bot]"`.

internal/github/comment_test.go
brushmtn: Fixed in 361a4a1 — rewrote as a table-driven test where each case sets `commentClient` and `appsClient` independently, pointed at its own isolated server. Confirmed with `go test -run TestGetCommentUserApp ./internal/github/...` in isolation.

internal/github/comment_test.go
brushmtn: Fixed in 361a4a1 — added `payload-app-empty-slug.json` (fallback to name) and `payload-app-empty.json` (both empty, expects error) as additional table-driven cases. Confirmed both new branches now show hit count 1 via `go test -coverprofile`.

internal/github/comment_test.go
```diff
@@ -386,3 +387,84 @@ func TestCommentNotHead(t *testing.T) {
...
+			appServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if r.URL.Path != "/app" {
+					t.Errorf("Mock server not configured to serve path %s", r.URL.Path)
+					w.WriteHeader(http.StatusNotFound)
+					return
+				}
+				payload, filePath, err := readFileToByteArray(tc.fixture)
+				if err != nil {
+					t.Fatalf("readFileToByteArray(%s) failed: %s", filePath, err)
+				}
```
vrivellino, line 421: Minor, non-blocking — `t.Fatalf` here runs on the HTTP handler's goroutine, not the test goroutine.

`Fatalf` calls `runtime.Goexit`, which [the `testing` docs](https://pkg.go.dev/testing#T.FailNow) require you to call from the goroutine running the test. On a handler goroutine it kills the handler mid-response instead of failing the test, so a missing or unreadable fixture would surface as an unexpected-EOF at the client rather than a clean `readFileToByteArray` failure.

Line 412 in this same handler already does it the right way with `t.Errorf`, so this looks like an oversight rather than intent. Matching it fixes the case:

```go
payload, filePath, err := readFileToByteArray(tc.fixture)
if err != nil {
    t.Errorf("readFileToByteArray(%s) failed: %s", filePath, err)
    w.WriteHeader(http.StatusInternalServerError)
    return
}
```

Worth noting `go vet` does not catch this — its `testinggoroutine` analyzer does not follow goroutines created via `httptest.NewServer`, so nothing in CI will flag it.

vrivellino (commented):
Re-reviewed `361a4a1` — all three findings are genuinely fixed, and I verified each rather than taking the replies at face value.

- **Finding 1:** guarding on `app == nil` plus validating the derived `appLogin` closes the `"[bot]"` hole. Dropping the widened pointer check was the right call — the derived check subsumes it.
- **Finding 2:** `go test -run TestGetCommentUserApp ./internal/github/...` now passes in isolation. I also ran `-shuffle=on` three times and `-count=2`; all green. The order dependency is gone, not masked.
- **Finding 3:** coverage confirms both new branches now execute (`comment.go:163.21,166.4` and `comment.go:167.21,170.4`, each hit count `1`, previously `0`). `getCommentUser` coverage went 70.7% → 77.3%.

`go build`, `gofmt`, `go vet`, `go test ./...` and `go test -race` are all clean. The table-driven rewrite is a better structure than what I originally suggested.

One minor nit inline — non-blocking, fine to merge with or without it.

internal/github/comment_test.go
brushmtn: Good catch — fixed in 5e86d32, matching the pattern on line 412: `t.Errorf` plus an explicit 500 response and early return.

vrivellino (approved):

PR #297, opened 2026-08-25
fix(github): make commentLogin caching concurrency-safe

### LINKED ISSUE #296: getCommentUser(): commentLogin caching is not concurrency-safe
`getCommentUser()` guards `commentLogin` with `mux` (an `RWMutex`), but the guarding is incomplete in two ways. Neither causes incorrect output today, so this is low priority — filing it so the seam is documented rather than rediscovered.

Found while reviewing #295. Pre-existing on `main`; not introduced by that PR.

## Where concurrency comes from

In webhook-server mode, `handleWebhook` spawns a goroutine per event:

```go
// internal/server/http_server.go:112
go process_event.ProcessCodeChange(eventInfo, wp.DevMode, &wp.Wg, &ignoredError)
```

Each of those reaches `getExistingComments()` → `getCommentUser()`. Two PR events arriving close together therefore run this function concurrently. Run-once and GitHub Actions modes are single-event, so they are unaffected.

## 1. Check-then-set is not atomic

`internal/github/comment.go:142-181`

```go
mux.RLock()
if commentLogin != "" {
    mux.RUnlock()
    return nil
}
mux.RUnlock()          // <-- lock released here
// ... GitHub API call ...
mux.Lock()
commentLogin = appLogin + "[bot]"
mux.Unlock()
```

The read lock is released before the API call. Two goroutines that both observe an empty `commentLogin` will both call `GET /app` (or `GET /user`), then both write.

Impact is limited: both goroutines compute the same value, so the cached result is correct either way. The cost is a redundant API call against the rate limit, only on the first events after startup.

## 2. Two reads of `commentLogin` take no lock at all

These reads are outside any `mux` critical section, so they can race with the `mux.Lock()` write above. This is a genuine data race that `go test -race` would flag if the package had a concurrent test:

- `internal/github/comment.go:182` — `log.Info().Msgf("Github Comment user name: %s", commentLogin)`
- `internal/github/comment.go:293` — `if isGithubAction || bypass || *c.User.Login == commentLogin {`

Line 293 is the consequential one. It is the comment-author match that decides whether argo-diff edits its existing PR comment or posts a new one. A torn or stale read there would produce a duplicate comment. In practice both writers store the same string, so a stale read can only mean "still empty" — which is already handled, since `getExistingComments()` calls `getCommentUser()` first and returns on error.

## Suggested fix

Collapse the whole populate-once path into a `sync.Once`, which removes the check-then-set gap outright, and add a small locked accessor for reads:

```go
var commentLoginOnce sync.Once

func getCommentLogin() string {
    mux.RLock()
    defer mux.RUnlock()
    return commentLogin
}
```

Then have line 182 and line 293 call `getCommentLogin()` instead of touching the variable directly.

A cheaper alternative is to hold the write lock across the API call. That serializes concurrent first-events on a network round trip, which is acceptable given it happens once per process, but `sync.Once` expresses the intent better.

## Acceptance

- [ ] `commentLogin` is populated exactly once per process, even under concurrent `getCommentUser()` calls.
- [ ] Every read of `commentLogin` holds `mux.RLock()` or goes through an accessor that does.
- [ ] A test exercising concurrent `getExistingComments()` passes under `go test -race`.

brushmtn:
Fixes #296: `getCommentUser()` released its read lock before calling the Github API and only took the write lock to set `commentLogin`, leaving a check-then-set gap where two concurrent webhook events could both miss the cache and both hit the API. Two other reads of `commentLogin` (the completion log line and the comment-author match in `getExistingComments()`) took no lock at all, racing with those writes under `go test -race`.

This holds the write lock across the whole check-call-set sequence in `getCommentUser()` and adds a locked `getCommentLogin()` accessor for the two previously-unlocked reads. Adds `TestGetCommentUserConcurrent`, a `-race` regression test that fires 50 concurrent `getCommentUser()` calls and asserts the Github API is hit exactly once.

No new dependencies. No behavior change on the success path; on the (rare) first-call-fails path, retries are still possible on the next event since `commentLogin` is only set after a successful API call.


brushmtn:

<details>
<summary>Plan</summary>

1. Confirm the race described in issue #296 by reading `internal/github/comment.go`.
2. Fix `getCommentUser()` to hold `mux.Lock()` across the whole check-then-call-then-set sequence, removing the check-then-set gap and the redundant duplicate API call.
3. Add a locked `getCommentLogin()` accessor and switch the two unlocked reads (log line, comment-author match in `getExistingComments()`) to use it.
4. Add a `go test -race` regression test (`TestGetCommentUserConcurrent`) that fires many concurrent `getCommentUser()` calls and asserts the mock API is hit exactly once.
5. Rebase onto latest `origin/main` (it moved during the session and had touched the same function for an unrelated app-slug feature), resolve the merge conflict, and re-verify build + `go test -race ./...`.
6. Push branch and open draft PR.

</details>

<details>

Done: fix implemented and committed (`internal/github/comment.go`, `internal/github/comment_test.go`), rebased onto latest `origin/main` after an unrelated upstream change touched the same function, conflict resolved, pushed.

Verified: `go build -v ./...` passes. `go test -race ./...` passes for the whole repo, including the new `TestGetCommentUserConcurrent` in `internal/github`. `gofmt -l .` reports nothing. `golangci-lint run ./internal/github/...` shows only pre-existing findings unrelated to this change (in `status.go` and the pre-existing test helpers), nothing new introduced by this diff.

Addressed maintainer's inline review comments: enabled `-race` in `.github/workflows/go.yml`'s test step (CI previously ran without it, so a regression of the locking pattern wouldn't have been caught), hoisted the `getCommentLogin()` read in `getExistingComments()` out of the pagination loop, and fixed `TestGetCommentUserConcurrent` to save/restore `commentClient`/`commentClientIsApp`/`commentLogin` (mirroring `TestGetCommentUserApp`) instead of leaking state into later tests. The fourth comment was informational only (about the PR description's wording), no code change needed.

Left to do: none — all three acceptance criteria from issue #296 are met (populate-once under concurrency, all reads locked, a passing `-race` concurrency test), and all actionable review feedback is addressed. `go build -v ./...` and `go test -race ./...` both pass. Ready for review.

</details>

internal/github/comment_test.go
```diff
@@ -367,6 +369,71 @@ func TestCommentExistingMultiNoComment(t *testing.T) {
 	}
 }
 
+// TestGetCommentUserConcurrent is the regression test for issue #296: concurrent
+// calls to getCommentUser() (as happen when the webhook server processes two PR
+// events at once) must populate commentLogin exactly once, and every read of it
+// must be race-free under `go test -race`.
```
vrivellino, line 375: **CI never runs `-race`, so half of this test's value is dead.**

`.github/workflows/go.yml:87` runs `go test -v ./...` with no `-race` flag.

The call-count assertion below still catches a regression of the main check-then-set bug — I confirmed it reports `got 50` against the old code even without the race detector. But the `getCommentLogin()` accessor exists *solely* to fix data races. If someone later reverts an accessor call back to a bare `commentLogin` read, CI will stay green and this test will pass.

Suggest changing that workflow line to:

```yaml
run: go test -race -v ./...
```

The package suite takes ~1.3s under `-race` locally, so the cost is negligible.

internal/github/comment_test.go
```diff
@@ -367,6 +369,71 @@ func TestCommentExistingMultiNoComment(t *testing.T) {
...
+			return
+		}
+		w.WriteHeader(http.StatusOK)
+		w.Write(payload)
+	}))
+	defer server.Close()
+
+	baseURL := server.URL + "/"
+	var err error
+	commentClient, err = github.NewClient(github.WithAuthToken("test1234"), github.WithURLs(&baseURL, &baseURL))
```
vrivellino, line 409: **This test mutates package globals without restoring them, and implicitly assumes `commentClientIsApp == false`.**

It sets `commentClient` and `commentLogin` but never restores either, and it never sets `commentClientIsApp`.

Compare `TestGetCommentUserApp` further down, which saves and restores all four globals in a `defer`. That asymmetry is the fragile part: this works today only because of source order (no earlier test flips `commentClientIsApp`). If a future test sets `commentClientIsApp = true` and forgets to restore it, this test silently takes the App branch and fails with a confusing `Mock server not configured to serve path /app` rather than anything about concurrency.

Suggest mirroring the sibling test's pattern:

```go
origCommentClient := commentClient
origIsApp := commentClientIsApp
origCommentLogin := commentLogin
defer func() {
	commentClient = origCommentClient
	commentClientIsApp = origIsApp
	mux.Lock()
	commentLogin = origCommentLogin
	mux.Unlock()
}()
commentClientIsApp = false
```

internal/github/comment.go
```diff
@@ -299,7 +307,7 @@ func getExistingComments(ctx context.Context, owner, repo string, prNum int) ([]
 		log.Debug().Msgf("Checking %d comments in %s/%s#%d", len(comments), owner, repo, prNum)
 		for _, c := range comments {
 			if strings.Contains(*c.Body, commentIdentifier) {
-				if isGithubAction || bypass || *c.User.Login == commentLogin {
+				if isGithubAction || bypass || *c.User.Login == getCommentLogin() {
```
vrivellino: **Nit: hoist this out of the loop.**

`getCommentLogin()` takes `RLock` once per matching comment, inside a loop that itself runs once per page of results.

This isn't a real cost — by the time control reaches here, `getCommentUser()` has already returned and the lock is uncontended (and when `isGithubAction || bypass` is true, short-circuiting means the accessor is never called at all). But reading it once above the pagination loop is cleaner and pins a single value for the whole scan:

```go
var login string
if !isGithubAction && !bypass {
	login = getCommentLogin()
}
```

then compare against `login` in the condition. Entirely optional.

internal/github/comment.go
```diff
@@ -184,9 +194,7 @@ func getCommentUser(ctx context.Context) error {
 			log.Error().Msg("Empty user returned - not sure how I got here")
 			return fmt.Errorf("empty user info")
 		}
-		mux.Lock()
 		commentLogin = *user.Login
-		mux.Unlock()
 	}
 	log.Info().Msgf("Github Comment user name: %s", commentLogin)
```
vrivellino, line 199: Informational only, no change requested — the PR description says the accessor fixed "the completion log line", but that isn't quite what happened.

This line still reads `commentLogin` directly. It's now race-free because `getCommentUser()` holds the write lock through to `return`, not because it goes through `getCommentLogin()`. The outcome is identical and the code is correct; only the description is slightly off.

Worth keeping in mind for whoever reads this line next: it is safe *only* while it stays inside the write-locked region. It must not be switched to `getCommentLogin()`, since `sync.RWMutex` isn't re-entrant and that would deadlock.

vrivellino (commented):
Reviewed the fix and confirmed it's correct — no correctness bugs found. I checked out the branch, ran the suite under `-race`, and verified the new test genuinely fails against the pre-fix code (`expected exactly 1 call to GET /user, got 50` plus `race detected during execution of test`).

Two things I specifically verified as *not* problems:
- **No re-entrancy deadlock.** The only read inside `getCommentUser()` is the `log.Info()` line, which reads `commentLogin` directly while already holding the write lock rather than going through the new accessor. Correct as written.
- **The lock-across-network-call is bounded.** Every path into `getCommentUser()` wraps the context in a timeout (`internal/server/http_server.go:182`, `internal/process_event/code_change.go:93`), so a hung GitHub API call can't stall comment processing indefinitely.

Four inline notes below — one worth acting on before merge (CI doesn't run `-race`), two minor, one purely informational.

internal/github/comment_test.go
brushmtn: Fixed — CI now runs `go test -race -v ./...` in `.github/workflows/go.yml`.

internal/github/comment_test.go
brushmtn: Fixed — mirrors `TestGetCommentUserApp`'s save/restore pattern now (`commentClient`, `commentClientIsApp`, `commentLogin`) so it no longer leaks state.

internal/github/comment.go
brushmtn: Applied — `login` is now read once via `getCommentLogin()` before the pagination loop and compared against inside it.

internal/github/comment.go
brushmtn: Good catch on the wording, thanks — no code change needed here, agreed this line is race-free because it's inside the write-locked region, not because of the accessor.

vrivellino (approved):

PR #302, opened 2026-08-29
feat(chart): add PodDisruptionBudget and Deployment rollout strategy support

TheRealNoob:
## Summary
- Add an optional `podDisruptionBudget` resource to the `argo-diff` Helm chart, so operators can protect pods from voluntary disruptions (node drains, cluster autoscaler)
- Add `deployment.strategy` to let operators override the Deployment's rollout strategy instead of relying on Kubernetes' default (RollingUpdate, 25%/25%)

## Details
- `templates/pdb.yaml` fails to render if `deployment.replicas == 1` (a PDB is either a no-op or blocks node drains forever at that replica count), and if `minAvailable`/`maxUnavailable` are both set (Kubernetes itself rejects a PDB spec with both fields populated)
- `deployment.replicas: 0` (scale-to-zero) is allowed — the PDB selector simply matches no pods
- New `tests/pdb_test.yaml` and additions to `tests/deployment_test.yaml` cover both features

## Test plan
- [x] `helm lint charts/argo-diff`
- [x] `helm unittest charts/argo-diff` (all suites pass)
- [ ] Reviewer: confirm chart version bump policy — none included here per `scripts/prep-release.sh` convention (bumped at release time, not per-PR)

Assisted-by: Claude Code (claude-opus-4-6)

### DIFF
--- a/charts/argo-diff/templates/pdb.yaml
+++ b/charts/argo-diff/templates/pdb.yaml
@@ -1,0 +1,32 @@
+{{- if .Values.podDisruptionBudget.enabled -}}
+{{- if eq (int .Values.deployment.replicas) 1 }}
+{{/* allow scaling down to replicas=0 without erroring */}}
+{{- fail "podDisruptionBudget.enabled requires deployment.replicas > 1" }}
+{{- end }}
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: {{ include "argo-diff.fullname" . }}
+  namespace: {{ include "argo-diff.namespace" . }}
+  labels:
+    {{- include "argo-diff.labels" . | nindent 4 }}
+  {{- with .Values.podDisruptionBudget.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{- if and .Values.podDisruptionBudget.minAvailable .Values.podDisruptionBudget.maxUnavailable }}
+  {{- fail "podDisruptionBudget.minAvailable and podDisruptionBudget.maxUnavailable are mutually exclusive; set only one" }}
+  {{- end }}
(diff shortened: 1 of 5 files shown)


charts/argo-diff/templates/pdb.yaml
```diff
@@ -0,0 +1,32 @@
...
+  name: {{ include "argo-diff.fullname" . }}
+  namespace: {{ include "argo-diff.namespace" . }}
+  labels:
+    {{- include "argo-diff.labels" . | nindent 4 }}
+  {{- with .Values.podDisruptionBudget.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{- if and .Values.podDisruptionBudget.minAvailable .Values.podDisruptionBudget.maxUnavailable }}
```
vrivellino, line 18: FYI: `0` is falsy in Go templates, so `maxUnavailable: 0` renders as `maxUnavailable: 1`, and `minAvailable: 0` silently falls through to the `maxUnavailable` branch. The mutual-exclusion check has the same hole — `minAvailable: 2` with `maxUnavailable: 0` passes it.

I'll accept your PR and make a correction in a quick follow-up of my own

vrivellino (approved):
@TheRealNoob Thank you for your contribution!

There a few smalls tweaks we'll need to make the chart before I can ship, but I'll handle that in a follow-up.

vrivellino:
@TheRealNoob I've released [chart v2.14.0](https://github.com/vince-riv/argo-diff/releases/tag/chart-2.14.0), which includes [v2.14.0](https://github.com/vince-riv/argo-diff/releases/tag/2.14.0). This latest chart includes this PR.

TheRealNoob:
Thank you!!!!

PR #306, opened 2026-09-01
docs(readme): document stacked-PR retargets in the Actions example

brushmtn:
A pull request stacked on another branch never ran argo-diff under the example GitHub Actions workflow. While its base was the parent branch, `branches: [main]` filtered its `opened`/`synchronize` events out; when the parent merged, GitHub retargeted it to `main` with a `pull_request` event carrying the `edited` action, which is not one of the default activity types. The check was skipped silently on exactly the pull requests whose base moves.

This updates the README's GitHub Actions example to add `edited` to `types`, gate the job with `if: github.event.action != 'edited' || github.event.changes.base.ref.from != ''` so title and body edits do not re-run it, and explain the behavior in a note. A second note records that the deployed webhook service still acts on `opened` and `synchronize` only, so retargets do not reach it there; commenting `argo diff` on the pull request is the manual fallback.

Documentation only — no code change and no new dependency.

Assisted-by: Claude Code (claude-opus-5)

PR #307, opened 2026-09-01
fix(webhook): handle stacked-PR base retargets in the deployed webhook path

brushmtn:
GitHub reports a stacked pull request's retarget onto `main` (after its parent branch merges) as a `pull_request` event with the `edited` action, not `synchronize`. `ProcessPullRequest()` in `internal/webhook/process.go` only accepted `opened` and `synchronize`, so the deployed webhook service silently dropped these events — the same gap already fixed for the GitHub Actions example workflow in #306, but on the always-on webhook path this time.

This adds a nil-safe check for `changes.base.ref.from`, GitHub's own signal that an `edited` event changed the base branch rather than the title or body, and accepts the event only in that case. No webhook re-configuration is needed: the existing "Pull requests" subscription already delivers `edited` events, GitHub Actions mode already handles this by not filtering on action type, and the pull request's current head/base are re-read from the API on every run.

Also includes two derived test fixtures and coverage in `process_test.go`, plus small doc/log fixes (`internal/webhook/context.md` had a stale `go-github/v89` reference; the ignore-log line printed the whole `Repository` struct instead of its full name).

No new dependency.


brushmtn:

<details>
<summary>Plan</summary>

1. Confirm go-github v90's `PullRequestEvent.Changes.Base.Ref.From` field and that its accessor chain is nil-safe, so it can be read without extra nil checks on a payload with no `changes` block.
2. Add a `baseRefChanged()` helper in `internal/webhook/process.go` and widen `ProcessPullRequest()`'s action filter to accept `edited` only when that helper is true.
3. Add two derived fixtures to `webhook_testdata/` (base retarget, title-only edit) built from the existing `payload-pr-sync.json`.
4. Extend `process_test.go`: fold both fixtures into the existing table-driven test, and add a dedicated test asserting the retargeted `BaseRef` differs from the old base.
5. Update `internal/webhook/context.md` (action-filter description, stale go-github version) and the README note that previously said the deployed service misses these retargets.
6. Verify: `go build ./...`, `go fmt ./...`, `go test ./...`, `golangci-lint run`.

</details>

<details>

**Status: complete, two review rounds addressed.**

Branch created fresh from `origin/main` after #306 merged (base commit `7794637`).

Round 1 (commit `59502ce`): added the base-retarget check to `ProcessPullRequest()`, two derived
fixtures, tests, and doc updates. Build/vet/fmt/test/lint all clean.

Round 2 review nits (commit `5f58a30`): inlined `baseRefChanged()` into `ProcessPullRequest()`
binding the accessor chain once, strengthened `TestPullRequestBaseRetarget` to assert the exact new
base value and dropped its duplicate checks, removed stale `before`/`after` keys from both derived
fixtures. Re-verified clean.

Round 3 review follow-up (commit `da4ecd9`): the round-2 refactor left `internal/webhook/context.md`
describing the now-deleted `baseRefChanged()` helper in two places, and its fixture recipe didn't
mention the `before`/`after` key removal from round 2. Fixed both references; `go test
./internal/webhook/...` and `golangci-lint run` re-verified clean.

Not done / out of scope: `reopened` is still not an accepted action in the webhook path (separate
gap). GitHub Actions mode needed no code change.

</details>

vrivellino (approved):

PR #311, opened 2026-09-02
feat: add `app-in-any-namespace` support

TheRealNoob:
Adds support for [app-in-any-namespace](https://argo-cd.readthedocs.io/en/latest/operator-manual/app-any-namespace/).

I took the approach of making it a mandatory argument since every Application CRD object has to live in a namespace.  Let me know if you have any concerns about the tests - happy to change however you want.

Oh and one thing I stumbled across worth mentioning.  This change adds the parameter `--app-namespace` to two commands:
* `argocd app diff` added support for `app-namespace` since v2.5.0.  So a while ago.
* `argocd app manifests` however only added support recently in v3.5.0 [#27942](https://github.com/argoproj/argo-cd/pull/27942).  Fortunately argo-diff [bundles](https://github.com/vince-riv/argo-diff/blob/d10994dd06d2e7115222f00e511e7eae3ac5ec61/Dockerfile#L12) v3.5.2 so it's in-scope now.  This hung me up on my local testing.

These versions above are for the argocd-cli.  The argocd-apiserver however has had support for namespace since at least v2.12.0 so we shouldn't have to bump the [minVersion](https://github.com/vince-riv/argo-diff/blob/d10994dd06d2e7115222f00e511e7eae3ac5ec61/internal/argocd/helper.go#L20) at all.

Assisted-by: Claude Code (claude-sonnet-4-6-20250514)


vrivellino:
This seems pretty straightforward and is certainly a feature that argo-diff should support.

I'm thinking through some of the implications, and the only notable one is that `argocd app manifests` is used for recursively detecting app-of-apps changes **and** argo-diff supports an operator to specify a specific argocd CLI path (in case someone wanted to pin the argocd CLI version). So, if someone were to specify a version of the argocd CLI that is <3.5 and they have app(s)-of-apps, argo-diff will fail.

Probably the most straightforward way to handle this is to gate use `--app-namespace` on `argocd app manifests` to client version >=3.5. This means the ArgoCD client version check will need to be pulled out of `ConnectivityCheck` and that any app-of-apps ArgoCD Applications will need to be in the `argocd` namespace (or wherever the user has deployed argocd).

This might be an edge case, but I want to protect backward compatibility.

internal/argocd/helper.go
```diff
@@ -577,7 +584,7 @@ func argoAppsWithChanges(ctx context.Context, appName string, appResources []App
 	}
 	// generate full manifests for our application at the specified revision
 	log.Debug().Msgf("argoAppsWithChanges(%s) - getting manifests at revision %s", appName, revision)
-	manifests, err := getApplicationManifests(ctx, appName, revision)
+	manifests, err := getApplicationManifests(ctx, appName, appNamespace, revision)
 	if err != nil {
 		log.Debug().Err(err).Msgf("argoAppsWithChanges() - getApplicationManifests(%s, %s) failed", appName, revision)
 		return argoApps, err
```
TheRealNoob: `getApplicationManifests` above returns an error when the Application exists in a non-default namespace and no `--app-namespace` is defined.  However this if statement block silences it.  No warnings or errors are reported to the github comment.  This logic looks to predate this PR.

TheRealNoob:
hmmm I didn't know that feature existed.  That makes this PR much more complicated 😞.  I spent a while thinking about it and decided that your proposed solution was best.  I would have preferred a solution where argo-diff could know ahead of time which `argocd app manifests` calls would succeed/fail based on the namespace, but that requires knowing what the default namespace is, but that isn't possible with just the current code.  The most straightforward way would be for the user to tell us, but that seems like a weird input for a niche edge-case.  

While talking with claude double checking everything to make sure there's no errors it found one that I thought was worth bringing up.  It seems pre-existing and this failure case exagerates it.  I'll drop claude's explanation below as I tried explaining it and it goes through too many calls for me to do a good job.

---

The `--app-namespace` version gate for `argocd app manifests` has a pre-existing gap worth flagging separately: some failures in this code path are only ever logged server-side, and never show up in the commit status or the PR comment.

**The gap:** when `argocd app manifests` fails while looking for nested `Application` children of a top-level app, that failure is silently absorbed unless it happens to coincide with the overall run timing out.

**Code path** (at [dd0f81b](https://github.com/TheRealNoob/argo-diff/commit/dd0f81b2ca1f4908a2989a0460acd2647bcdbd35)):

1. [`getApplicationManifests`](https://github.com/TheRealNoob/argo-diff/blob/dd0f81b2ca1f4908a2989a0460acd2647bcdbd35/internal/argocd/argocd_client.go#L261-L278) runs `argocd app
manifests` and returns whatever error the CLI produced.
2. [`argoAppsWithChanges`](https://github.com/TheRealNoob/argo-diff/blob/dd0f81b2ca1f4908a2989a0460acd2647bcdbd35/internal/argocd/helper.go#L568-L591) passes that error straight up.
3. [`processTopLevelApp`](https://github.com/TheRealNoob/argo-diff/blob/dd0f81b2ca1f4908a2989a0460acd2647bcdbd35/internal/argocd/helper.go#L175-L186) is where it gets absorbed:
   ```go
   appsWithChanges, err := argoAppsWithChanges(ctx, app.Name, app.Namespace, appResChanges.ChangedResources, eventInfo.Sha)
   if err != nil {
       if ctx.Err() != nil {
           res.notDiffed = append(res.notDiffed, fmt.Sprintf("nested apps of %s", app.Name))
       }
       log.Warn().Err(err).Msgf("Unable to determine if argo app %s has other argo apps with changes", app.Name)
       return res
   }
   ctx.Err() only reports whether the overall processing deadline (ARGO_DIFF_TIMEOUT) has been exceeded; it says nothing about the CLI's own exit code. An ordinary, fast CLI failure
(bad flag, app not found, RBAC denied) leaves ctx.Err() nil, so this branch never records the app as skipped — it just logs a warning and moves on.
4. In https://github.com/TheRealNoob/argo-diff/blob/dd0f81b2ca1f4908a2989a0460acd2647bcdbd35/internal/process_event/code_change.go#L167-L208, only WarnStr (set when a top-level app's own diff fails) and notDiffed (set on timeout) affect the reported outcome. Since this failure sets neither, the app is reported as a plain success — its own diff renders fine, and its nested children are just quietly never checked.

Why this matters for the version gate specifically: on a pinned CLI < 3.5.0, any app-of-apps parent living outside the CLI's default namespace will now reliably hit this path — argocd app manifests can't find it once --app-namespace is omitted — and the result is a silently incomplete diff rather than any visible warning.

This isn't new. The same "log and continue" pattern predates this PR, and an identical gap exists one level deeper in
https://github.com/TheRealNoob/argo-diff/blob/dd0f81b2ca1f4908a2989a0460acd2647bcdbd35/internal/argocd/helper.go#L213-L230, already flagged there via a comment pointing at #273. I'm raising it here because this fix makes the pre-existing gap realistically triggerable (old pinned CLI + non-default-namespace app-of-apps).

internal/argocd/helper.go
TheRealNoob: I believe this failure case impacts scenarios beyond just this one that i'm highlighting

vrivellino:
@TheRealNoob I started #314 so that we can add a notice/advisory string to app diffs. eg: for multi-namespace feature in this PR, we could add a notice that argo-diff is unable to discover children of app-of-apps. IMO that would be sufficient if users are pinning an old version of the argocd cli. While I'm in there, I'm taking a look at the overall design/style of diff comments. Once #314 lands, I think you can incorporate it here and get this merged.

vrivellino:
> ... While I'm in there, I'm taking a look at the overall design/style of diff comments. Once #314 lands, I think you can incorporate it here and get this merged.

@TheRealNoob You can see a preview of the style refactor here: https://github.com/vince-riv/argo-diff/pull/314#issuecomment-5609915311

vrivellino:
@TheRealNoob Just re-reviewed this now that we have a way to post a notice to users who may have the argocd cli pinned to old versions. I'm going to have my agent post the results. I think we have a way forward after you merge update with main

Also, I can cut a pre-release if you'd like to deploy with all of your changes. I appreciate the contributions!

internal/argocd/helper.go
```diff
@@ -165,7 +172,7 @@ func processTopLevelApp(ctx context.Context, app Application, appLookup map[stri
 		return res
 	}
 	res.diffResult = &appResChanges
-	appsWithChanges, err := argoAppsWithChanges(ctx, app.Name, appResChanges.ChangedResources, eventInfo.Sha)
+	appsWithChanges, err := argoAppsWithChanges(ctx, app.Name, app.Namespace, appResChanges.ChangedResources, eventInfo.Sha)
```
vrivellino, line 181: bug: `appLookup` (`helper.go:301`) is keyed by `app.Name` alone, but `argocd app list` returns Applications from every namespace, so two apps named `web` in `team-a` and `team-b` collapse into one entry. `appLookup[subApp.Name]` at line 197 can then hand `getMultiSrcAppChanges()` the wrong namespace's Application, and `diffApplication()` posts a diff for the wrong app.

internal/argocd/argocd_client.go
```diff
@@ -158,6 +168,60 @@ func argocdVersion(ctx context.Context) (string, string, error) {
...
+		return *cachedSupportsManifestsAppNamespace
+	}
+	v, err := argocdClientVersion(ctx)
+	if err != nil {
+		log.Warn().Err(err).Msg("failed to determine argocd CLI client version; omitting --app-namespace from `argocd app manifests` (app-of-apps discovery will only find nested apps in the CLI's default namespace)")
+		return false
+	}
+	supported := versionAtLeast(v, appNamespaceManifestsMinVersion)
+	cachedSupportsManifestsAppNamespace = &supported
+	return supported
```
vrivellino, line 230: suggestion: `config.AddNotice()` landed in #314 for exactly this case — `internal/config/context.md` calls it "the seam for that first version-gated feature," and nothing calls it yet. Raise one in both `false` branches, so a reader whose app-of-apps children are missing learns a pinned CLI caused it instead of getting only ArgoCD's opaque "permission denied" via the per-app `NoticeStr`.

```suggestion
	v, err := argocdClientVersion(ctx)
	if err != nil {
		log.Warn().Err(err).Msg("failed to determine argocd CLI client version; omitting --app-namespace from `argocd app manifests` (app-of-apps discovery will only find nested apps in the CLI's default namespace)")
		config.AddNotice("argo-diff could not determine the `argocd` CLI version, so app-of-apps children are only discovered for Applications in ArgoCD's own namespace.")
		return false
	}
	supported := versionAtLeast(v, appNamespaceManifestsMinVersion)
	if !supported {
		config.AddNotice(fmt.Sprintf(
			"The pinned `argocd` CLI (%s) is older than %s, so app-of-apps children are only discovered for Applications in ArgoCD's own namespace. Point `ARGOCD_CLI_CMD_NAME` at a newer CLI to diff nested Applications in other namespaces.",
			v, appNamespaceManifestsMinVersion))
	}
	cachedSupportsManifestsAppNamespace = &supported
	return supported
```

Needs `"github.com/vince-riv/argo-diff/internal/config"` added to this file's imports — no cycle, `internal/config` imports nothing internal. `AddNotice()` dedupes, so the uncached error branch is safe to call on every request.

README.md
```diff
@@ -240,7 +240,7 @@ the accepted environment variables and their respective GitHub Actions inputs.
 | APP_ENV                          | N/A                         | no               |          | Set to `dev` during local development. |
 | ARGOCD_AUTH_TOKEN                | argocd_auth_token           | yes              |          | Bearer token for ArgoCD (value passed to `--auth-token`). |
 | ARGOCD_APP_DIFF_SERVER_SIDE_DIFF | argocd_app_server_side_diff | no               |          | Set `--server-side-diff` for `argocd app diff` (`true`/`false`). |
-| ARGOCD_CLI_CMD_NAME              | N/A                         | no               | `argocd` | Overrides the `argocd` CLI command name (e.g., to use a specific argocd version); accepts either a bare command name on `PATH` or an absolute path to the binary. |
+| ARGOCD_CLI_CMD_NAME              | N/A                         | no               | `argocd` | Overrides the `argocd` CLI command name (e.g., to use a specific argocd version); accepts either a bare command name on `PATH` or an absolute path to the binary.  Support for diffing Applications in a non-default namespace (most commonly `argocd`) requires CLI version >= 3.5.0. |
```
vrivellino: nit: `argocd` is the default namespace, so "(most commonly `argocd`)" contradicts "a non-default namespace". The 3.5.0 floor also applies only to app-of-apps discovery — `argocd app diff` has accepted `--app-namespace` since 2.5.0.

```suggestion
| ARGOCD_CLI_CMD_NAME              | N/A                         | no               | `argocd` | Overrides the `argocd` CLI command name (e.g., to use a specific argocd version); accepts either a bare command name on `PATH` or an absolute path to the binary. Discovering app-of-apps children of an Application that lives outside ArgoCD's own namespace (most commonly `argocd`) requires CLI version >= 3.5.0. |
```

internal/argocd/argocd_client.go
```diff
@@ -209,14 +277,14 @@ func getApplicationManifests(ctx context.Context, appName, revision string) ([]K
...
 }
 
-func diffApplication(ctx context.Context, appName string, revision string, revisions []string, srcPos []int) ([]AppResource, error) {
+func diffApplication(ctx context.Context, appName string, appNamespace string, revision string, revisions []string, srcPos []int) ([]AppResource, error) {
 	var appResList []AppResource
 	log.Trace().Msg("diffApplication() called")
 	// argocd app diff argo-diff --revision XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX [--refresh]
 	// argocd app diff argo-diff --revisions XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX --source-positions 1 --revisions a.b.c --source-positions 2
-	args := []string{"app", "diff", appName, "--revision", revision}
+	args := []string{"app", "diff", appName, "--app-namespace", appNamespace, "--revision", revision}
```
vrivellino, line 293: nit: this is unconditional on the grounds that `minVersion` (2.12.0) already clears `app diff`'s 2.5.0 floor, but `ConnectivityCheck()` is skippable via `ARGO_DIFF_BYPASS_CONNECTIVITY_CHECKS=argocd`. An operator who bypasses the check and also pins a pre-2.5 CLI through `ARGOCD_CLI_CMD_NAME` gets an unknown-flag failure on every diff.

EDIT: I don't care too much about this nit. I can't imagine someone pinning pre-2.5 at this point in time!

vrivellino (commented):
One blocking finding, plus a suggestion now that #314 has merged.

`processTopLevelApp()` on main already turns a failed app-of-apps enumeration into a `NoticeStr`, so the version gate no longer drops children silently once this branch rebases. What's missing is the *reason*: the reader gets ArgoCD's opaque "permission denied" with nothing explaining that a pinned CLI caused it. `config.AddNotice()` (new in #314, still with no callers) closes that.

- `internal/argocd/helper.go:613` — `argoAppsWithChanges()` matches rendered manifests to changed resources by `app.Name` alone (`slices.Contains(argoAppNamesFound, name)`). Same name-keying gap as `appLookup`, and `AppResource` already carries the namespace to disambiguate.
- This branch needs a rebase onto main. #314 edited the same `processTopLevelApp()` error branch this PR touches, so `helper.go` will conflict.

internal/argocd/helper.go
TheRealNoob: Fixed in f39b9cf. `appListToMap` now keys by `namespace/name` via a new `appKey()` helper, and `processTopLevelApp` looks up by `appKey(subApp.Namespace, subApp.Name)`. Also fixed the same class of bug in `multiSrcAppNamesDiffed` (wave-3 dedup), which had the identical name-only collision. Test fixtures updated to set `Namespace: "argocd"` on all test Applications.

internal/argocd/argocd_client.go
TheRealNoob: Applied in c15760f, closely following your suggestion. One wording change: replaced "discovered" with "enumerated" to match the existing `NoticeStr` message in `processTopLevelApp` ("Unable to enumerate the app-of-apps children of...").

README.md
TheRealNoob: Applied in c15760f, using your suggested wording verbatim.

internal/argocd/argocd_client.go
TheRealNoob: Acknowledged — leaving this as-is per your note. Agreed that pinning pre-2.5 is not a realistic scenario at this point.

TheRealNoob:
Rebased onto main and addressed all review feedback in two new commits:

**f39b9cf — fix: disambiguate Application lookups by namespace/name**
- Added `appKey(namespace, name)` helper that returns `namespace/name`
- `appListToMap` now keys by `appKey` instead of `app.Name` alone
- `processTopLevelApp`'s `appLookup` lookup uses `appKey`
- `argoAppsWithChanges` matches changed resources by `appKey` instead of name alone
- `multiSrcAppNamesDiffed` (wave-3 dedup) also converted — same class of bug, caught during review
- Test fixtures updated to set `Namespace: "argocd"` on all test Applications and diff headers

**c15760f — feat: surface version-gate notices via config.AddNotice()**
- `supportsManifestsAppNamespace()` now calls `config.AddNotice()` in both false-return branches (version lookup failed, and version too old), closely following your suggested diff
- One wording change: "discovered" → "enumerated" to match the existing `NoticeStr` in `processTopLevelApp`
- README `ARGOCD_CLI_CMD_NAME` wording updated per your suggestion

The `argoAppsWithChanges` name-matching bug from the review body (helper.go:613) is fixed as part of the first commit — same `appKey` treatment applied there.

`go build`, `go test -race`, `go vet`, `gofmt`, and `golangci-lint run` (Docker, v2.13.2) all pass.

Assisted-by: Claude Code (claude-sonnet-4-6-20250514)

vrivellino (approved):

PR #313, opened 2026-09-08
feat: add user-definable refresh comment trigger

TheRealNoob:
Allows customization of the keyword(s) that trigger an event refresh.  Should be fully backwards compatible.  Despite the number of files modified it really is a quite simple change.

One of the things I've been looking at is the current chart's inability to allow users to define their own inputs to the program, either via ENV VARs or args.  So if you'd rather I implement that now and remove the configmap additions I can do that.


action.yml
```diff
@@ -68,6 +68,10 @@ inputs:
     description: 'Max number of ArgoCD applications diffed concurrently (capped at 32). Defaults to 4'
     required: false
     default: ''
+  refresh_comment_keywords:
+    description: 'Comma-separated list of PR comments that re-trigger argo-diff, replacing the default "argo diff,argo-diff"'
+    required: false
+    default: ''
```
vrivellino, line 74: This input can never take effect: `eventInfoFromEnv()` rejects any `GITHUB_EVENT_NAME` other than `pull_request` (`internal/server/run_once.go:85`), and `IsRefreshComment()` is reached only from the webhook server's `issue_comment` dispatch.

The github actions execution of argo-diff doesn't support comments at this time. My initial thought when I implemented it was that folks can retrigger the actions workflow from the Github UI. Supporting comment-driven re-triggers could be a future improvement, but I think we can leave it out of this PR.

Remove this input and  `ARGO_DIFF_REFRESH_COMMENT_KEYWORDS` env line below it, and mark the README's Actions-input column `N/A`, as `ARGO_DIFF_DISABLE_NON_GITHUB_REPO_MATCH` already does.

internal/github/comment.go
```diff
@@ -19,22 +19,33 @@ import (
...
+	refreshCommentKeywordsRaw := strings.TrimSpace(os.Getenv("ARGO_DIFF_REFRESH_COMMENT_KEYWORDS"))
+	if refreshCommentKeywordsRaw == "" {
+		refreshCommentKeywordsRaw = "argo diff,argo-diff"
+	}
+	for _, keyword := range strings.Split(refreshCommentKeywordsRaw, ",") {
+		keyword = strings.ToLower(strings.TrimSpace(keyword))
+		if keyword != "" {
+			refreshCommentKeywords = append(refreshCommentKeywords, keyword)
+		}
+	}
```
vrivellino, line 48: This is an edge case, but when `ARGO_DIFF_REFRESH_COMMENT_KEYWORDS=","` it passes the empty check on line 40, then every split element is dropped. `refreshCommentKeywords` stays nil and `IsRefreshComment()` is permanently false, so refresh comments die silently. Fall back to the defaults when nothing parses.

```suggestion
	}
	if len(refreshCommentKeywords) == 0 {
		log.Warn().Msg("ARGO_DIFF_REFRESH_COMMENT_KEYWORDS parsed to zero keywords; using defaults")
		refreshCommentKeywords = []string{"argo diff", "argo-diff"}
	}
```

internal/github/comment.go
```diff
@@ -19,22 +19,33 @@ import (
...
 func init() {
 	commentClientIsApp = false
 	mux = &sync.RWMutex{}
 	isGithubAction = os.Getenv("ARGO_DIFF_CI") != "true" && os.Getenv("GITHUB_ACTIONS") == "true"
 	contextStr = strings.TrimSpace(os.Getenv("ARGO_DIFF_CONTEXT_STR"))
+	refreshCommentKeywordsRaw := strings.TrimSpace(os.Getenv("ARGO_DIFF_REFRESH_COMMENT_KEYWORDS"))
+	if refreshCommentKeywordsRaw == "" {
+		refreshCommentKeywordsRaw = "argo diff,argo-diff"
+	}
+	for _, keyword := range strings.Split(refreshCommentKeywordsRaw, ",") {
```
vrivellino, line 43: [nit] This parsing sits in init(), so a test can only reach it by setting the env var before package load; comment_test.go covers neither it nor IsRefreshComment(). Extract parseRefreshCommentKeywords(raw string) []string and table-test it.

internal/github/comment.go
```diff
@@ -118,16 +129,16 @@ func ConnectivityCheck() error {
...
-	if input == "argo diff" || input == "argo-diff" {
-		return true
-	}
-	if input == "argo diff "+strings.ToLower(contextStr) {
-		return true
-	}
-	if input == "argo-diff "+strings.ToLower(contextStr) {
-		return true
+	for _, keyword := range refreshCommentKeywords {
+		if input == keyword || input == keyword+" "+strings.ToLower(contextStr) {
```
vrivellino, line 139: [nit] `strings.ToLower(contextStr)` is recomputed for every keyword on every comment event. Lowercase it once in init() into a package-level var

vrivellino:
> One of the things I've been looking at is the current chart's inability to allow users to define their own inputs to the program, either via ENV VARs or args. So if you'd rather I implement that now and remove the configmap additions I can do that.

Does issue #317 and proposed solution capture this?

vrivellino:
@TheRealNoob: Thanks for adding this feature! I left a few in-line comments; the one on `action.yml` is blocking

action.yml
TheRealNoob: Fixed — removed the `refresh_comment_keywords` input and its `ARGO_DIFF_REFRESH_COMMENT_KEYWORDS` env mapping from `action.yml`, and updated the README env var table's Actions-input column to `N/A` with a note that this only takes effect in webhook-server deployment mode.

internal/github/comment.go
TheRealNoob: Fixed — added a `len(keywords) == 0` check after parsing with a `log.Warn()` fallback to the defaults. The parsing is also now extracted into `parseRefreshCommentKeywords()` (see below), which made this path easy to cover in a table test.

internal/github/comment.go
TheRealNoob: Done — extracted into `parseRefreshCommentKeywords(raw string) []string` and added `TestParseRefreshCommentKeywords` covering: empty → defaults, whitespace-only → defaults, only-commas → defaults, single custom keyword, multiple keywords with trimming/lowercasing, and empty entries between commas dropped. Also added `TestIsRefreshComment` for the matching logic.

internal/github/comment.go
TheRealNoob: Done — added `lowerContextStr` to the package-level `var` block, assigned once in `init()`, and `IsRefreshComment` now uses it directly.

TheRealNoob:
Thank you.  All should be addressed.  Ready for re-review.

vrivellino (approved):

PR #314, opened 2026-09-09
refactor: redesign PR comments, add a top-level notice, and fix the comment size budget

vrivellino:
Adds an advisory (`NoticeStr`) channel for the case where an application's diff succeeded but something alongside it degraded, then reworks the PR comment around it: the design, the way severity is signalled, and the comment size budget.

## Why

Three problems compounded each other.

**Severity was invisible.** The fatal `WarnStr` and the advisory `NoticeStr` were passed through the same `ArgoAppMarkdown.WarnStr` field and rendered as an identical unlabeled fence, told apart only by an `"Error: "` prefix the caller pasted on. A reader could not tell "this diff is untrustworthy" from "this diff is fine, but FYI".

**The design did not scale.** ASCII `===` decoration, app names title-cased by `capitalizeWords()` into something no longer copy-pasteable, status split across three stacked lines, and every `<details>` forced `open` — so a change matching fifteen applications was an unnavigable wall.

**The size budget leaked.** `String()` checked `len(md+candidate)` against `maxCommentLen`, but `Comment()` then wrapped *every* body with an unbounded operator preamble (`ARGO_DIFF_COMMENT_PREAMBLE`) plus the HTML marker — entirely outside that budget. The gap between `maxCommentLen` (261500) and GitHub's real 262144 was 644 bytes to absorb all of it. A long preamble plus a full-size body is a 422 from the API and **no comment at all**.

## Severity is now carried by the alert type

| Field | Alert | Colour | Meaning |
| --- | --- | --- | --- |
| `CommentMarkdown.Notices` | `> [!NOTE]` | blue | comment-level advisory, from `config.Notices()` |
| `ArgoAppMarkdown.NoticeStr` | `> [!NOTE]` | blue | this app's diff is good, something alongside it degraded |
| `ArgoAppMarkdown.ErrStr` | `> [!CAUTION]` | red | this app's diff **failed**; no resources are shown |
| `timeoutMarkdown()` | `> [!WARNING]` | yellow | applications left undiffed |

`ErrStr` and `NoticeStr` are separate fields on an `AppMarkdownOpts` struct now, so severity is a field rather than a string prefix.

## Style changes

**Application header — before:**

```
▼ === Grafana ===
https://argocd.example.com/applications/argocd/grafana
Synced :white_check_mark:
Healthy :green_heart:
```

**After** — one scannable line, the raw app name, and a change count:

```
▼ grafana · 2 changed · ✅ Synced · 💚 Healthy
   Open in ArgoCD ↗
```

**Resource header — before:**

```
▼ ===== apps/Deployment grafana/grafana =====
```

**After** — the size of the change is visible before you expand it:

```
▼ apps/Deployment grafana/grafana · +2 −2
```

**Advisory vs fatal — before**, indistinguishable:

````
```
Unable to discover app-of-apps children of meta: rpc error: ...
```

```
Error: Failed to diff application broken-deployment: ...
```
````

**After**, a blue note above diffs that still matter, and a red caution where none are shown:

````
---

> [!NOTE]
> **meta** - Unable to discover app-of-apps children of meta: rpc error: ...

<details open>
<summary><b>meta</b> · 1 changed · ⚠️ OutOfSync · ⏳ Progressing</summary>
...
</details>

---

> [!CAUTION]
> **broken-deployment** - diff failed, so no manifests are shown. ❓ Unknown · 💚 Healthy
>
> ```
> Failed to diff application broken-deployment: ...
> ```
>
> [Open in ArgoCD ↗](...)
````

### Why the alerts sit outside the block

**GitHub renders its coloured alert boxes only at the top level of a document.** Inside any HTML element — `<details>` included — the extension is skipped and the block degrades to a plain blockquote whose first line is a literal `[!NOTE]`. Extra blank lines, a blank line before `<summary>`, a `<div>` wrapper and nesting inside another blockquote were all tried against GitHub's own renderer and all degrade the same way.

So an application's alerts are emitted between the `---` rule and its `<details>`. Three consequences: a hoisted alert names its application, since it is detached from the block; an application with an error and no resources renders as **just its alert**, because there are no diffs to fold; and `appOpen()` never folds an application carrying a notice or an error, so diffs an alert calls into question are not hidden behind a fold.

The failure is invisible from Go — the emitted string looks perfectly correct, and nothing validates it at runtime. The guard is a test helper: `checkBodyWellFormed()` in `markdown_test.go` walks `<details>` depth under `go test` and fails on any `> [!` line nested inside one. Every splitting test runs it on every body.

Other changes: emoji are literal Unicode rather than `:shortcodes:` (shortcode substitution is not reliable inside raw HTML `<summary>`); blocks fold automatically once a comment covers more than 3 applications or an application has more than 5 changed resources; a summary index table lists every application; and continuation bodies carry a `part i of n` header instead of opening as a naked wall of diffs.

## Budget

- `commentWrapperLen()` is `len(wrapComment(""))` — derived from the same function `Comment()` posts with, so the budget and the wrapper cannot drift apart.
- `splitReserve` covers every string appended *after* a fit check has passed: the closing `</details>`, the continuation marker, the part header.
- A resource's cap is derived from the budget instead of the hardcoded `maxResourceDiffLen`, so a lowered `ARGO_DIFF_COMMENT_MAX_CHARS` can never leave a diff that no body can hold.
- `ErrStr`, `NoticeStr` and each notice are bounded to 4000 bytes; `HealthMsg` to 500. All come from unbounded input (`err.Error()`, operator config, the cluster).
- A mid-application split now closes the app's `<details>` before flushing — it used to leave it unbalanced.
- `finalize()` hard-truncates any body still over budget. A 422 means no comment at all, which is worse than a truncated one.

## New options

| Environment Variable | Input | Default | What it does |
| --- | --- | --- | --- |
| `ARGO_DIFF_COMMENT_NOTICE` | `comment_notice` | | Advisory text at the top of the comment. Several notices separated by `\|` |
| `ARGO_DIFF_COMMENT_INDEX_COUNT` | `comment_index_count` | `2` | When to render the index: `-1` always, `0` never, any other `n` at `n` applications |
| `ARGO_DIFF_COMMENT_COLLAPSE` | `comment_collapse` | `auto` | `auto`, `expanded` or `collapsed` |
| `ARGO_DIFF_COMMENT_MAX_CHARS` | `comment_max_chars` | `262144` | Comment size cap; cannot be raised past GitHub's own limit |

All four land in `README.md`, `action.yml`, `charts/argo-diff/values.yaml`, the configmap template and its unittests, `docs/k8s/`, and `run_once.go`'s env log. `config.commentIndexCount` is the one key emitted without a plain `{{- with }}`: `0` is meaningful for it and Helm treats a numeric `0` as unset.

`config.AddNotice()` is the in-process seam for the same channel. Nothing calls it yet — `ConnectivityCheck()` already parses the argocd client and server versions and discards them, which is where the first version-gated advisory would hook in.

## Tests

`markdown.go` had no test for splitting, the budget, or truncation. It does now, asserting invariants rather than golden output: every body balances its `<details>` tags and code fences, no alert is ever nested inside a `<details>`, and every body stays under the cap **once wrapped** — that last one fails on `main`.

Two bugs surfaced during the build and are fixed: a whitespace-only notice rendered as an empty alert, and a low `ARGO_DIFF_COMMENT_MAX_CHARS` collapsed every diff into the "too large" marker instead of splitting across bodies.

Review found two more, both fixed in `c873a1c`: `maxResourceBodyLen()` did not reserve the hoisted alerts (~8KB) or the `---` rule, so an application carrying an alert alongside a near-cap resource overshot the budget; and `finalize()`, the declared last guard, cut at an arbitrary byte and left the ````diff```` fence and two `<details>` open, with the truncation marker stranded inside the fence. `finalize()` now closes what it cuts. Reverting either fix reproduces the malformed body at the 180-changed-line threshold, so both are covered by tests.

Also removed: `unknownCount`, which was declared and reported but never incremented, and a stale `GITHUB_STATUS_CONTEXT_STR` in `docs/k8s/deployment.yaml` that the Go code stopped reading when `ARGO_DIFF_CONTEXT_STR` replaced it.

`go test ./...`, `go test -race`, `golangci-lint run` (0 issues), `helm lint` and `helm unittest` (59 tests) all pass.

## Live preview

Everything below this line is real output from the current renderer, verified through GitHub's own `POST /markdown`: three alert boxes, zero literal `[!NOTE]` markers.

**3 of 15 apps with changes** compared to live state
<sub>1:47AM UTC, 18 Dec 2023</sub>

> [!NOTE]
> `argocd` CLI 2.12 is in use; server-side diff needs 2.14 or newer, so it is disabled for this run.

| Application | Changed | Sync | Health |
| --- | --: | --- | --- |
| [grafana](https://argocd.example.com/applications/argocd/grafana) | 2 | ✅ Synced | 💚 Healthy |
| [meta](https://argocd.example.com/applications/argocd/meta) | 1 | ⚠️ OutOfSync | ⏳ Progressing |
| [broken-deployment](https://argocd.example.com/applications/argocd/broken-deployment) | ❗ | ❓ Unknown | 💚 Healthy |

---

<details open>
<summary><b>grafana</b> · 2 changed · ✅ Synced · 💚 Healthy</summary>

[Open in ArgoCD ↗](https://argocd.example.com/applications/argocd/grafana)

<details open>
<summary><code>ConfigMap</code> grafana/grafana · <b>+2 −2</b></summary>

```diff
--- live.yaml
+++ a9eabee.yaml
@@ -26,9 +26,9 @@
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/name: grafana
-    app.kubernetes.io/version: 10.1.2
+    app.kubernetes.io/version: 10.2.2
     argocd.argoproj.io/instance: grafana
-    helm.sh/chart: grafana-6.60.2
+    helm.sh/chart: grafana-7.0.17
```

</details>

<details open>
<summary><code>apps/Deployment</code> grafana/grafana · <b>+1 −1</b></summary>

```diff
--- live.yaml
+++ a9eabee.yaml
@@ -87,7 +87,7 @@
 spec:
-  replicas: 1
+  replicas: 2
   revisionHistoryLimit: 10
```

</details>

</details>

---

> [!NOTE]
> **meta** - Unable to discover app-of-apps children of meta: rpc error: code = Unknown desc = repository not accessible

<details open>
<summary><b>meta</b> · 1 changed · ⚠️ OutOfSync · ⏳ Progressing</summary>

[Open in ArgoCD ↗](https://argocd.example.com/applications/argocd/meta)

<sub>waiting for rollout to finish</sub>

<details open>
<summary><code>argoproj.io/Application</code> argocd/helm-deployment · <b>+1 −1</b></summary>

```diff
--- live.yaml
+++ new.yaml
@@ -143,10 +143,10 @@
     path: charts/test-basic-deployment
-    targetRevision: k3s-test
+    targetRevision: main
```

</details>

</details>

---

> [!CAUTION]
> **broken-deployment** - diff failed, so no manifests are shown. ❓ Unknown · 💚 Healthy
>
> ```
> Failed to diff application broken-deployment: app diff broken-deployment --revision 496cddb: exit status 20: {"level":"fatal","msg":"rpc error: code = Unknown desc = test/does-not-exist: app path does not exist"}
> ```
>
> [Open in ArgoCD ↗](https://argocd.example.com/applications/argocd/broken-deployment)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


internal/github/markdown.go
```diff
@@ -2,221 +2,577 @@ package github
...
+}
+
+// maxResourceBodyLen is the largest a single resource's diff may render: what
+// is left of one body once this application's continuation header and the
+// resource's own markup are accounted for. Deriving it from the budget rather
+// than hardcoding it is what stops a lowered ARGO_DIFF_COMMENT_MAX_CHARS from
+// leaving a resource no body can hold.
+func (a ArgoAppMarkdown) maxResourceBodyLen(summary string) int {
+	const markup = 64 // the <details>/<summary> scaffolding around the body
+	n := commentBudget() - splitReserve - len(a.OverviewStr(true, false)) - len(summary) - markup
```
brushmtn: **Bug (high): this reservation is missing the app's alerts and the `---` rule, so a body can exceed the budget.**

`String()` emits `appSeparator + a.alerts() + OverviewStr(...)` ahead of the first resource, but only `OverviewStr` is subtracted here. `alerts()` is bounded by `maxNoticeLen` on each of `ErrStr` and `NoticeStr`, so it can add ~8 KB that nothing accounts for. The result overshoots `commentBudget()` and falls through to `finalize()`, which truncates through open markup (see the comment there).

Two smaller points in the same expression:

- `OverviewStr(true, false)` is the *continued* form. The first body renders `OverviewStr(false, true)`, which is the larger of the two — `<details open>` is 5 bytes longer than `<details>`. The 64-byte `markup` slack absorbs it today, but it is reserving the wrong one.
- `appSeparator` (6 bytes) is unreserved too.

Suggested:

```go
func (a ArgoAppMarkdown) maxResourceBodyLen(summary string) int {
	const markup = 64 // the <details>/<summary> scaffolding around the body
	// reserve the *first*-body header, which is the larger of the two, plus
	// everything String() emits ahead of the block: the rule and the alerts
	head := len(appSeparator) + len(a.alerts()) + len(a.OverviewStr(false, true))
	n := commentBudget() - splitReserve - head - len(summary) - markup
	if n < minResourceLen {
		return minResourceLen
	}
	return n
}
```

`ErrStr`/`NoticeStr` are set in `AppMarkdown()` before any `AddResourceDiff()` call, so `a.alerts()` is available here.

internal/github/markdown.go
```diff
@@ -2,221 +2,577 @@ package github
...
-	md += syncString(a.SyncStatus) + "\n"
-	md += healthString(a.HealthStatus, a.HealthMsg) + "\n\n"
-	if a.WarnStr != "" {
-		md += "```\n" + a.WarnStr + "```\n\n"
+	for i, b := range bodies {
+		if len(b) <= budget {
+			continue
+		}
+		log.Warn().Msgf("comment body %d is %d bytes, over the %d byte budget - truncating", i+1, len(b), budget)
+		bodies[i] = truncateBytes(b, budget-len(truncatedMarker)) + truncatedMarker
```
brushmtn: **Bug (high): the last guard is not markup-aware, so it produces exactly the malformed output the rest of `String()` works to prevent.**

`truncateBytes()` cuts at an arbitrary byte offset. When the cut lands inside a resource — which is where it lands in practice, since resources are the bulk of a body — it leaves the ```` ```diff ```` fence open and both the resource's and the application's `<details>` unclosed. `truncatedMarker` is then appended *inside* the open fence, so the one signal the reader gets renders as diff text.

Reproduced (see the review body): `open=2 close=0 fences=1`.

The comment above is right that a truncated comment beats a 422. But truncation should close what it cuts:

```go
for i, b := range bodies {
	if len(b) <= budget {
		continue
	}
	log.Warn().Msgf("comment body %d is %d bytes, over the %d byte budget - truncating", i+1, len(b), budget)
	t := truncateBytes(b, budget-len(truncatedMarker)-truncTailReserve)
	// close whatever the cut left open, innermost first, or the marker and
	// everything below it is swallowed
	if strings.Count(t, "```")%2 != 0 {
		t += "\n```\n"
	}
	t += truncatedMarker
	for range strings.Count(t, "<details") - strings.Count(t, "</details>") {
		t += detailsClose
	}
	bodies[i] = t
}
```

`truncTailReserve` needs to cover the closing fence plus a bounded number of `</details>` (nesting is at most 2 here). Fixing `maxResourceBodyLen()` above makes this path much harder to reach, but it is the declared last guard — it should still be safe on its own.

internal/github/markdown.go
```diff
@@ -2,221 +2,577 @@ package github
...
-					md += r
-				}
+			md += head
+			continue
+		}
+		overview := head + a.OverviewStr(false, appOpen)
+		resOpen := c.resourceOpen(a)
+		// look ahead to the first resource: opening an application block at the
+		// tail of a body that can't hold any of its diffs just wastes a header
+		if len(md)+len(overview)+len(a.Resources[0].render(resOpen)) > limit {
```
brushmtn: **Nit (low): this look-ahead has no progress guard, so it can emit a near-empty body.**

The resource loop below guards its split with `placed > 0`, and the comment there explains exactly why: a body holding only a header has no more room than a fresh one, so splitting cannot help. The same reasoning applies here, but there is no guard — when `overview + first resource` cannot fit in *any* body, `flush()` still fires and emits whatever the preamble happens to be.

That is the 57-byte `**Argo-Diff** - part 1 of 2\n**1 of 1 apps with changes**\n` body in the repro. Suggested:

```go
// only split if a fresh body would actually help - otherwise this just
// emits the preamble as its own body and still doesn't fit
if len(md)+len(overview)+len(first) > limit && len(overview)+len(first) <= limit {
	flush()
}
```

brushmtn (commented):
## Re-review

Re-reviewed after the refactor, against `b21c90b..5cd3011`. Verified locally: `go build ./...`, `go test ./...`, `go test -race`, `golangci-lint run` (0 issues), `gofmt -l`, `helm lint` and `helm unittest` (59 tests) all pass. The summary's claims about the wrapper-budget leak, the alert-nesting failure, and the four options reaching every deployment surface all check out.

**One real bug, reproduced locally: `finalize()` can emit a body with two unclosed `<details>` tags and an unclosed ```` ```diff ```` fence.** Details inline on `maxResourceBodyLen()` and `finalize()`.

### Repro

`ARGO_DIFF_COMMENT_MAX_CHARS=20000`, one app with a 4000-byte `NoticeStr`, one resource whose diff has 200 changed lines:

```
body 0: len=   57   <- "part 1 of 2" plus the preamble, nothing else
body 1: len=19979   open=2  close=0  fences=1
```

Body 1 ends:

```
+  key175: new value padding padding padding
-  key176: ol

`<<< TRUNCATED - comment size limit reached >>>`
```

Three consequences: two `<details>` never close, the ```` ```diff ```` fence never closes, and the truncation marker lands *inside* that open fence, so a reader sees it as diff text rather than as a notice. The threshold is 180 changed lines at that cap — 150 fits, 180 breaks.

At the default 262144 cap this needs a single ~255 KB resource diff on an app that also carries a notice: rare, but reachable. At any lowered `ARGO_DIFF_COMMENT_MAX_CHARS` — an option this PR adds and documents — it is easy.

`checkBodyWellFormed()` is exactly the guard that catches this, but no test drives an app carrying **both** an alert and a near-cap resource. Worth adding that case alongside the fix.

### Note on the summary

> `checkBodyWellFormed()` therefore walks `<details>` depth and fails on any `> [!` line nested inside one, which is the guard that would have caught this.

`checkBodyWellFormed()` lives in `markdown_test.go` and runs only under `go test`. The sentence reads as though the renderer checks itself at runtime — worth rewording so the distinction is clear, since this is the section a future reader will come back to.

### What is good

- The wrapper-budget fix is real, and deriving `commentWrapperLen()` from the same `wrapComment()` that posts is the right shape — the two genuinely cannot drift.
- The alert-hoisting fix is correct and non-obvious. GitHub does skip the alert extension inside `<details>`, and the failure is invisible from Go.
- Tests assert invariants rather than golden output, which is the right call for a renderer.
- All four options land in `README.md`, `action.yml`, `values.yaml`, the configmap template, its unittests, `docs/k8s/` and `run_once.go`. The non-`with` emit for `commentIndexCount` is correctly justified.
- `unknownCount` and the stale `GITHUB_STATUS_CONTEXT_STR` are both correct dead-code removals.

Marked as a comment rather than a change request — the bug is worth fixing before merge, but nothing here is a blocker on the design.

internal/github/markdown.go
```diff
@@ -2,221 +2,577 @@ package github
...
 	}
+	return b.String()
+}
+
+// diffStats counts added and removed lines in a unified diff, skipping the
+// ---/+++ file headers so they don't read as one added and one removed line.
+func diffStats(diffStr string) (added, removed int) {
+	for line := range strings.SplitSeq(diffStr, "\n") {
+		switch {
+		case strings.HasPrefix(line, "+++"), strings.HasPrefix(line, "---"):
```
brushmtn: **Nit (low): this also skips removed YAML document separators, undercounting the `−N` badge.**

A removed `---` separator renders in a unified diff as `----` (the `-` prefix plus the three dashes), which matches `HasPrefix(line, "---")` and is skipped. Multi-document YAML embedded in a ConfigMap is the common case.

Cosmetic — it only affects the `+N −M` summary badge, not the diff itself. Matching the headers exactly is enough:

```go
case strings.HasPrefix(line, "+++ "), strings.HasPrefix(line, "--- "):
```

That still misses a removed `--- ` document separator with a trailing comment, but the `argocd` CLI emits its headers as `--- live.yaml` / `+++ <sha>.yaml`, so keying on the space is a strict improvement.

internal/argocd/helper.go
```diff
@@ -173,6 +173,12 @@ func processTopLevelApp(ctx context.Context, app Application, appLookup map[stri
 			// one entry anyway: without it the run reports as complete while
 			// every nested application diff is missing.
 			res.notDiffed = append(res.notDiffed, fmt.Sprintf("nested apps of %s", app.Name))
+		} else {
+			// Not a deadline: this app's own diff is good and already in
+			// res.diffResult, only its app-of-apps children are missing. That's
+			// advisory, not fatal - notDiffed would report it as a timeout and
+			// fail the run, and WarnStr would suppress the diff we do have.
+			appResChanges.NoticeStr = fmt.Sprintf("Unable to discover app-of-apps children of %s: %s", app.Name, err.Error())
```
brushmtn: **Design question, carried over from the first review and still open.**

`argoAppsWithChanges()` returns an error only *after* it has confirmed `argoAppNamesFound` is non-empty — i.e. after it knows this app's diff contains changed `argoproj.io/Application` resources. So a non-timeout failure here is not "we could not check whether there are children". It is "there are children with changes, and we cannot diff any of them."

That is the same partial result as the deadline branch directly above, which deliberately fails the run. `internal/process_event/context.md` states the rule:

> Reporting success on a partial diff is worse than failing.

This branch routes the identical incompleteness to `StatusSuccess` and a zero exit in Actions mode.

Concretely: `argocd app manifests` returns `rpc error: repository not accessible`, the parent's own diff renders, every child app diff is silently absent from the comment, and the check is green.

Two coherent resolutions:

1. Keep it advisory, but say what is missing. The current text — "Unable to discover app-of-apps children of X" — does not tell the reader that diffs are absent from the comment. Something like "…; diffs for its child Applications are missing from this comment" would.
2. Treat it like `notDiffed` and fail, matching the deadline branch.

The choice is deliberate and documented, so this is a judgement call rather than a defect — but it is worth an explicit answer either way.

internal/process_event/code_change.go
```diff
@@ -173,15 +173,26 @@ func ProcessCodeChange(eventInfo webhook.EventInfo, devMode bool, wg *sync.WaitG
...
+			_ = cMarkdown.AppMarkdown(github.AppMarkdownOpts{
+				Name: appName, ErrStr: a.WarnStr,
+				SyncStatus: appSyncStatus, HealthStatus: appHealthStatus, HealthMsg: appHealthMsg,
+			})
 			if firstError == "" {
 				firstError = a.WarnStr
 			}
 		} else {
 			log.Trace().Msgf("%s has %d Changed Resources", appName, len(a.ChangedResources))
 			if len(a.ChangedResources) > 0 {
```
brushmtn, line 187: **Nit (low): `NoticeStr` is read only inside this branch, so it is dropped silently if the invariant ever changes.**

The comment further down is correct that today `NoticeStr` implies `changeCount > 0` — `processTopLevelApp()` returns early on `len(ChangedResources) == 0` before it can set the field. But that invariant lives in another package, and if a future producer sets `NoticeStr` on an app with no changed resources, the advisory vanishes with no trace.

A `log.Warn()` in the `else` is enough to make that loud:

```go
} else if a.NoticeStr != "" {
	log.Warn().Msgf("%s has a NoticeStr but no changed resources - dropping it: %s", appName, a.NoticeStr)
}
```

internal/argocd/helper_test.go
```diff
@@ -376,6 +377,75 @@ func TestGetApplicationChangesOutOfTimeEnumeratingNestedApps(t *testing.T) {
 	if !slices.Equal(notDiffed, want) {
 		t.Errorf("notDiffed = %v, want %v", notDiffed, want)
 	}
+	// the deadline path stays on notDiffed; NoticeStr is for the non-timeout
+	// failure only, so reporting can't describe a real error as a timeout or
+	// vice versa
+	if len(appResList) == 1 && appResList[0].NoticeStr != "" {
```
brushmtn: **Nit (low): this assertion can disappear instead of failing.**

The length check above it uses `t.Errorf`, not `t.Fatalf`, so if `len(appResList)` ever regresses the test reports that one failure and this `NoticeStr` check silently stops running — the guard turns a second failure into no signal at all.

Promote the earlier check to `t.Fatalf` and drop the guard:

```go
if len(appResList) != 1 {
	t.Fatalf("expected the parent app's diff in the results, got %d results", len(appResList))
}
...
if appResList[0].NoticeStr != "" {
	t.Errorf("NoticeStr = %q, want empty when the deadline is what stopped nested app enumeration", appResList[0].NoticeStr)
}
```

internal/github/markdown.go
```diff
@@ -2,221 +2,605 @@ package github
...
-	md += syncString(a.SyncStatus) + "\n"
-	md += healthString(a.HealthStatus, a.HealthMsg) + "\n\n"
-	if a.WarnStr != "" {
-		md += "```\n" + a.WarnStr + "```\n\n"
+	for i, b := range bodies {
+		if len(b) <= budget {
+			continue
+		}
+		log.Warn().Msgf("comment body %d is %d bytes, over the %d byte budget - truncating", i+1, len(b), budget)
+		t := truncateBytes(b, budget-len(truncatedMarker)-truncTailReserve)
```
brushmtn, line 615: **Verified — the repair holds at every budget I tried.** Balanced `<details>`, paired fences, marker outside the fence. Reverting it reproduces the original `open=2 close=0 fences=1`.

One degenerate case remains, entirely optional:

```
budget=20 -> len=51, over=true
```

`truncateBytes()` returns `""` for a non-positive `max`, and then ~50 bytes of `truncatedMarker` go on regardless. So when `budget < len(truncatedMarker)+truncTailReserve` (~77), every body collapses to a marker that is itself over budget.

Reaching it needs `ARGO_DIFF_COMMENT_MAX_CHARS` set below the operator's own `ARGO_DIFF_COMMENT_PREAMBLE` length, so **there is no 422 risk** — the only cap breached is one they set themselves. Worth closing only if you want the invariant to be unconditional:

```go
func commentBudget() int {
	n := commentMaxLen() - commentWrapperLen()
	if n < minResourceLen {
		log.Warn().Msgf("ARGO_DIFF_COMMENT_MAX_CHARS %d leaves only %d bytes after the preamble and marker", commentMaxLen(), n)
		return minResourceLen
	}
	return n
}
```

Your call — I would not hold the PR for it.

internal/argocd/context.md
```diff
@@ -116,6 +116,27 @@ application. When the deadline hits while enumerating an app-of-apps' children,
...
+- **`NoticeStr` — advisory.** The diff succeeded and `ChangedResources` is good, but something
+  alongside it degraded. `process_event` renders it above the app's diffs and the run still
+  succeeds.
+
+The one producer today is `processTopLevelApp()`: when `argoAppsWithChanges()` fails while `ctx`
+still has time left, the parent's own diff is already in `res.diffResult`, but its app-of-apps
+children are gone. Note **which** children: `argoAppsWithChanges()` returns `nil` early when it
+finds no changed `argoproj.io/Application` resources, so its only error path is reached *after*
+confirming there are children with changes. The notice therefore says their diffs are missing from
+the comment — "couldn't discover them" on its own reads as "there were none". That is `NoticeStr`, not `notDiffed` (which reporting renders as
```
brushmtn: **Nit (doc): this line is 144 characters.**

The new sentence joined onto the existing line rather than re-wrapping, so the paragraph now has one line at 144 chars where the rest of the prose sits near 100. The long lines elsewhere in this file (11-14) are table rows, which is different.

Re-wrapping the paragraph is all it needs — the content is right, and the "children *with changes*" point is exactly the fact that makes the notice wording matter.

brushmtn (commented):
## Re-review of `c873a1c` — both fixes verified

Verified locally on `c873a1c`: `go build ./...`, `go test ./...`, `go test -race`, `golangci-lint run` (0 issues) and `gofmt -l` all pass. **Your repro no longer reproduces, and I could not find a new way to break it.** Replies on each original thread; two small new things inline.

### The original repro, re-run

Same setup — `ARGO_DIFF_COMMENT_MAX_CHARS=20000`, one app with a 4000-byte `NoticeStr`:

```
lines= 150 body 0: len=17605 open=2 close=2 fences=2
lines= 180 body 0: len= 4306 open=2 close=2 fences=0   <- was open=2 close=0 fences=1
lines= 400 body 0: len= 4306 open=2 close=2 fences=0
lines=2000 body 0: len= 4308 open=2 close=2 fences=0
```

At 180 lines the resource now degrades to `<<< DIFF TOO LARGE TO DISPLAY >>>` rather than overflowing. That is the right outcome: the alert genuinely costs ~8 KB of resource budget now, and a single resource has never split across bodies.

I also drove a harsher case than the new test does — both alerts at `maxNoticeLen`, a 900-byte `HealthMsg`, a UI URL, a 500-byte preamble and four near-cap resources. Balanced, and under the cap once wrapped.

### `finalize()`'s repair, exercised directly

Called with a body ending mid-fence inside two open `<details>`:

```
budget=  100 -> open=1 close=1, marker present
budget=  200 -> open=2 close=2 fences=2
budget= 5000 -> open=2 close=2 fences=2
```

Balanced at every budget, with the marker outside the fence. A cut landing inside a partial `</details` also comes out balanced. One degenerate case left — inline on `finalize()`.

### The design question

The decision record in `internal/process_event/context.md` is the right answer. It states the position, the reason, and the concrete cost of revisiting it — including that `timeoutMarkdown()`'s "ran out of time" wording would need its own block. Putting the "children *with changes*" fact in `internal/argocd/context.md`, next to the code that depends on it, is where it belongs.

Nothing blocking from me.

internal/github/markdown.go
brushmtn: **Fixed and verified.** `head` now reserves `appSeparator + a.alerts() + OverviewStr(false, true)`, exactly as proposed, including both smaller points.

Confirmed by re-running the repro: at 180 changed lines the resource degrades to `<<< DIFF TOO LARGE TO DISPLAY >>>` instead of overflowing the body. That is the correct outcome — the alert really does cost ~8 KB of resource budget now, and a single resource has never split across bodies.

`TestAlertPlusLargeResourceStaysWellFormed` covers it, and I checked a harsher case than the test does (both alerts at `maxNoticeLen`, a 900-byte `HealthMsg`, a UI URL, a 500-byte preamble, four near-cap resources) — balanced, and under the cap once wrapped.

internal/github/markdown.go
brushmtn: **Fixed and verified.** Details in my new comment on the current `finalize()` — the repair holds at every budget I exercised, with the marker outside the fence.

`truncTailReserve = len(truncFenceClose) + 2*len(detailsClose)` is the right bound: the appends total exactly what is reserved, and two is the deepest nesting this renderer produces.

internal/github/markdown.go
brushmtn: **Fixed.** Taken as proposed — the second term is the same progress guard the resource loop already carried. The 57-byte "part 1 of 2" body is gone from the repro output.

internal/github/markdown.go
brushmtn: **Fixed.** Keying on the trailing space is right. The `argocd` CLI emits `--- live.yaml` / `+++ <sha>.yaml`, so the headers still match and a removed `----` document separator now counts toward the `−N` badge.

internal/argocd/helper.go
brushmtn: **Answered, and the wording is fixed.** The new text says plainly that the child diffs are missing from the comment, which is what a reader needs.

Resolution 1 is a reasonable call, and the reasoning holds: this is an ArgoCD-side repository problem, so failing the check would ask the PR author to fix something outside their change. What makes it reviewable is that `internal/process_event/context.md` now records it as a **decision** — the position, the reason, and the concrete cost of revisiting it, including that `timeoutMarkdown()`'s "ran out of time" wording would need its own block. That is the part that matters if someone reopens this later.

Putting the "children *with changes*" fact in `internal/argocd/context.md`, next to the code it constrains, is the right home for it.

internal/process_event/code_change.go
brushmtn: **Fixed.** The `else if` logs a warning, and the comment records that the invariant it depends on lives in another package. That is exactly the failure mode this guards.

internal/argocd/helper_test.go
brushmtn: **Fixed.** `t.Fatalf` on the count, guard dropped. A regression in the result count can no longer turn a second failure into no signal.

internal/github/markdown.go
```diff
@@ -2,221 +2,619 @@ package github
...
+// that needs ARGO_DIFF_COMMENT_MAX_CHARS set below the operator's own
+// ARGO_DIFF_COMMENT_PREAMBLE, so the only cap breached is one they set
+// themselves - but the warning is the part they need, since nothing else says
+// their preamble has eaten the comment.
+func commentBudget() int {
+	n := commentMaxLen() - commentWrapperLen()
+	if n < minResourceLen {
+		log.Warn().Msgf("ARGO_DIFF_COMMENT_MAX_CHARS %d leaves only %d bytes after the preamble and marker - using %d",
+			commentMaxLen(), n, minResourceLen)
+		return minResourceLen
```
brushmtn: **The floor closes the case I raised, but it opens a narrower one: it can raise the budget above what actually fits, and post a body GitHub rejects.**

When `0 < commentMaxLen()-commentWrapperLen() < minResourceLen`, this returns 512 — more than the room that is genuinely left. `wrapComment()` then adds the preamble on top.

Measured at the **default** cap with a 262000-byte `ARGO_DIFF_COMMENT_PREAMBLE`:

```
pre-floor  (c873a1c):  budget=  121   worst wrapped=262127   422=false
post-floor (12317a8):  budget=  512   worst wrapped=262285   422=true
```

`githubCommentHardMax` is 262144. The old code produced a useless 121-byte body but **posted successfully**; the new code gets a 422 and posts nothing.

**The root cause is a level up, and it is pre-existing.** `commentPreamble` is unbounded operator input, and it is now the only input to the size budget that nothing caps — `ErrStr`, `NoticeStr`, each notice, `HealthMsg` and every resource body are all bounded. `README.md` says "Keep to 150 chars or less", but nothing enforces it. At a 300000-byte preamble *both* versions 422, because the wrapper alone exceeds the limit; `markdown.go` cannot rescue that from here.

Bounding the preamble in `comment.go`'s `init()` closes both the regression and the pre-existing hole, and leaves this floor unconditionally safe:

```go
commentPreamble = strings.TrimSpace(os.Getenv("ARGO_DIFF_COMMENT_PREAMBLE"))
if len(commentPreamble) > maxPreambleLen {
	log.Warn().Msgf("ARGO_DIFF_COMMENT_PREAMBLE is %d bytes - truncating to %d", len(commentPreamble), maxPreambleLen)
	commentPreamble = truncateBytes(commentPreamble, maxPreambleLen)
}
```

With `commentWrapperLen()` bounded, `commentMaxLen()-commentWrapperLen()` can never fall into that band at the default cap, and the floor only ever engages on a deliberately low `ARGO_DIFF_COMMENT_MAX_CHARS` — which is what it was written for.

The alternative, if you would rather keep the change local, is to clamp the floor against the hard max here:

```go
if room := githubCommentHardMax - commentWrapperLen(); room < minResourceLen {
	return max(room, 0)
}
return minResourceLen
```

That fixes the regression but leaves the 300000-byte case unfixable, so I would prefer bounding the preamble.

**Severity is low** — this needs a ~256 KB preamble against a documented 150-char guideline. Flagging it because `githubCommentHardMax`'s own comment states the invariant without qualification: *"Nothing this package emits may exceed it."* Right now one construction does.

brushmtn (commented):
## Re-review of `12317a8`

The floor fix works — `finalize()` now stays inside the budget at every value I exercised, and the doc re-wrap is clean. Build, `go test ./...`, `go test -race`, `golangci-lint run` (0 issues) and `gofmt -l` all pass; all 10 CI checks are green.

**One new thing: the floor makes a 422 reachable where it was not before.** Narrow band, absurd configuration, low severity — but it is the exact outcome the budget design exists to prevent, so it is worth a look. Inline on `commentBudget()`.

### Floor fix verified

```
finalize budget= 512 len= 512 over=false open=2 close=2 fences=2
cap=  20 body 0 len=483 budget=512 over=false
cap=  77 body 0 len=483 budget=512 over=false
cap=1200 body 0 len=748 budget=777 over=false
```

The `len=51, over=true` case is gone, bodies stay well-formed, and none come back empty. `TestAbsurdlyLowCapStillProducesUsableBodies` picks the right boundary values.

The warning is the right call. An operator whose preamble has eaten the comment gets told, which is the part nothing else surfaced.

### Doc re-wrap verified

`internal/argocd/context.md:134` is gone. The only 100+ character lines left in that file are table rows and lines that pre-date this branch. Splitting the paragraph at the new sentence reads better than the original single block did.

internal/github/markdown.go
brushmtn: **Fixed and verified.** `commentBudget()`'s floor makes this invariant unconditional — `finalize()` stays inside the budget at every value I exercised, including at the floor itself:

```
finalize budget= 512 len= 512 over=false open=2 close=2 fences=2
finalize budget= 600 len= 600 over=false open=2 close=2 fences=2
```

The `len=51, over=true` case is gone, and low-cap bodies come back well-formed and non-empty. `TestAbsurdlyLowCapStillProducesUsableBodies` picks the right boundary values (`20`, `77`, `300`, `500`).

The warning was the right thing to add — nothing else told an operator their preamble had eaten the comment.

One consequence of the floor is worth its own thread: it can now raise the budget above what actually fits. See my new comment on `commentBudget()`.

internal/argocd/context.md
brushmtn: **Fixed.** The 144-character line is gone, and the only 100+ character lines left in the file are table rows and lines that pre-date this branch.

Splitting the paragraph at the new sentence reads better than the original single block did, and taking the neighbouring test paragraph with it was the right call.

internal/github/markdown.go
brushmtn: **Fixed and verified.** The clamp gives back only the room that genuinely remains. At the exact case I measured — default cap, 262000-byte wrapper:

```
                        budget   worstWrapped   422
12317a8 (before):          512         262288   yes
442305d (now):             121         262115   no
```

`boundPreamble()` closes it at the source too. Through `init()` — what production actually gets — no preamble reaches this path at all:

```
envPreamble=   150 -> bounded= 150  budget=261947  422=false
envPreamble=262000 -> bounded=4000  budget=258097  422=false
envPreamble=300000 -> bounded=4000  budget=258097  422=false
```

**Keeping both the bound and the clamp is right, and the stated reason holds.** I confirmed the clamp is load-bearing: a 258000-byte `ARGO_DIFF_CONTEXT_STR` yields `budget=4096` and stays under the limit, because `commentIdentifier` is unbounded by design.

Past roughly 262200 bytes of context string the identifier alone exceeds GitHub's cap, so `commentBudget()` returns 0 and nothing is postable. That is genuinely unfixable without bounding `contextStr`, which would break comment matching — and `Comment()` already logs the failure at error level, so an operator is not left guessing. Correctly out of scope.

`TestBudgetFloorNeverExceedsTheHardMax` picks the right four cases, and asserting *whether the floor engages* rather than just the resulting number is what makes it a real regression test.

One small thing came out of reading the new `init()` path — separate comment on `comment.go`.

internal/github/comment.go
```diff
@@ -35,9 +35,9 @@ func init() {
 	mux = &sync.RWMutex{}
 	isGithubAction = os.Getenv("ARGO_DIFF_CI") != "true" && os.Getenv("GITHUB_ACTIONS") == "true"
 	contextStr = strings.TrimSpace(os.Getenv("ARGO_DIFF_CONTEXT_STR"))
-	commentPreamble = strings.TrimSpace(os.Getenv("ARGO_DIFF_COMMENT_PREAMBLE"))
+	commentPreamble = boundPreamble(strings.TrimSpace(os.Getenv("ARGO_DIFF_COMMENT_PREAMBLE")))
 	if commentPreamble == "" {
-		commentPreamble = contextStr
+		commentPreamble = boundPreamble(contextStr)
```
brushmtn: **Nit (low): on this fallback path the warning names a variable the operator never set.**

`boundPreamble()` hardcodes its message:

```go
log.Warn().Msgf("ARGO_DIFF_COMMENT_PREAMBLE is %d bytes - truncating to %d", len(s), maxPreambleLen)
```

But the input here is `contextStr`, i.e. `ARGO_DIFF_CONTEXT_STR`. An operator with a long context string and no `ARGO_DIFF_COMMENT_PREAMBLE` at all gets told to go fix a variable they never configured — and since the preamble silently falls back to the context string, that is exactly the person least likely to make the connection.

Passing the name in keeps the message honest on both paths:

```go
func boundPreamble(name, s string) string {
	if len(s) <= maxPreambleLen {
		return s
	}
	log.Warn().Msgf("%s is %d bytes - truncating the comment preamble to %d", name, len(s), maxPreambleLen)
	return truncateBytes(s, maxPreambleLen)
}
```

Called as `boundPreamble("ARGO_DIFF_COMMENT_PREAMBLE", ...)` above and `boundPreamble("ARGO_DIFF_CONTEXT_STR", contextStr)` here.

This is the only thing I found in `442305d`. Everything from the three earlier reviews is fixed and verified — `go test -count=1 ./...`, `-race`, `golangci-lint` (0 issues), `gofmt`, `helm unittest` (59 tests) all pass, and all 10 CI checks are green.

vrivellino:
Thanks — everything across the four reviews is addressed in `c873a1c`, `12317a8`, `442305d` and `23ee2b4`. Two things worth saying out loud rather than leaving in the thread replies.

**The design question.** Keeping an app-of-apps enumeration failure advisory rather than failing the run is a deliberate call, and your reading of the code is what made it a real question: `argoAppsWithChanges()` only errors after confirming there are children *with changes*, so this genuinely is a partial diff. The reasoning for staying green is that this is an ArgoCD-side repository problem — failing the check would ask a PR author to fix something outside their change. That is arguable, so `internal/process_event/context.md` records it as a decision with its cost, including that `timeoutMarkdown()`'s "ran out of time" wording would need its own block if it is ever revisited.

**The floor was my regression.** Good catch on `commentBudget()`. Fixing your `len=51` case introduced a worse one — the floor returned more headroom than remained, turning a useless-but-postable 121-byte body into a 422. Bounding the preamble was the right call: it was the last input to the budget that nothing capped. I also kept the local clamp against `githubCommentHardMax`, since `commentIdentifier` carries `ARGO_DIFF_CONTEXT_STR` and that stays unbounded — comment matching depends on its exact value.

internal/github/comment.go
brushmtn: **Fixed and verified.** Taken as suggested — the name is an argument, and both call sites pass the variable the value actually came from.

Threading the name through rather than, say, branching inside `boundPreamble()` is the right shape: the function has no way to know which path called it, and the doc comment now says why the argument exists rather than just what it is.

Verified on `23ee2b4`: `go build ./...`, `go test -count=1 ./...`, `go test -race`, `golangci-lint run` (0 issues), `gofmt -l`, `helm lint` and `helm unittest` (59 tests) all pass, and all 10 CI checks are green.

This closes the last open item. I have re-checked every finding from all four reviews against this head commit and confirmed each fix is present, so I am resolving the threads.

PR #315, opened 2026-09-10
feat: add ARGO_DIFF_REQUIRE_APP_MATCH flag

TheRealNoob:
Deployed org-wide, argo-diff posts a commit status to every PR the webhook receives, before it knows whether ArgoCD even tracks the repo. In large orgs this creates a lot of noise on PRs ArgoCD doesn't care about; leading to questions from users.

This PR adds an opt-in flag `ARGO_DIFF_REQUIRE_APP_MATCH`.  It does a cheap match check `argocd app list` + filtering.  If it finds no results, then it doesn't post anything to the PR: no pending or success.  When something does match, behavior is unchanged.

Trade-offs (documented in README and `internal/process_event/context.md`):
- One extra `argocd app list` call, and slower first feedback, on PRs that do match.
- A PR that previously matched and later stops matching won't have its stale comment cleared under this flag. An explicit refresh comment (`argo diff`) will still trigger a run and update the comment.

Assisted-by: Claude Code (claude-sonnet-4-6)


vrivellino:
@TheRealNoob Thanks again for yet another contribution! :-) This is a great feature to add. My original line of thought it adding the status check to every repo when deployed org-wide is that one might want to have it as a required check. But adding an opt-out is totally reasonable.

I an agent do a code review, and I'm going to have it post in-line comments. I'd also like to rename the feature flag env var to `ARGO_DIFF_REQUIRE_APP_MATCH`. IMHO it's a little more clear of the intention.

vrivellino:
Also - if you merge in main or rebase, #316 should quiet the failing docker build check

vrivellino (commented):
Inline comments below; verdict and context are in the PR thread.

internal/process_event/code_change.go
```diff
@@ -129,6 +129,21 @@ func ProcessCodeChange(eventInfo webhook.EventInfo, devMode bool, wg *sync.WaitG
...
 
+	// ARGO_DIFF_SKIP_UNMATCHED_EVENTS: check for a matching ArgoCD application before posting commit status
+	if strings.ToLower(strings.TrimSpace(os.Getenv("ARGO_DIFF_SKIP_UNMATCHED_EVENTS"))) == "true" {
+		matched, err := argocd.HasMatchingApplications(ctx, eventInfo)
+		if err != nil {
+			log.Error().Err(err).Msg("argocd.HasMatchingApplications() failed")
+			_ = github.Status(ctx, github.StatusError, err.Error(), eventInfo.RepoOwner, eventInfo.RepoName, eventInfo.Sha, devMode)
+			*callerErr = err
+			return
+		}
```
vrivellino, line 140: bug: this `github.Status` call runs on `ctx`, which is the very context whose expiry is the most likely cause of the error it is reporting. A hung `argocd app list` burns the whole `ARGO_DIFF_TIMEOUT`, `HasMatchingApplications` returns `context.DeadlineExceeded`, and this call is cancelled before it reaches GitHub, so the PR ends up with no status at all and the failure is swallowed by `_ =`. Every other reporting call in this function deliberately runs on a `context.Background()`-derived context for exactly this reason (see "Timeout budget" in `internal/process_event/context.md`).

```suggestion
		if err != nil {
			log.Error().Err(err).Msg("argocd.HasMatchingApplications() failed")
			statusCtx, statusCancel := context.WithTimeout(context.Background(), reportReserve(timeout))
			defer statusCancel()
			_ = github.Status(statusCtx, github.StatusError, err.Error(), eventInfo.RepoOwner, eventInfo.RepoName, eventInfo.Sha, devMode)
			*callerErr = err
			return
		}
```

internal/argocd/helper.go
```diff
@@ -388,6 +388,35 @@ func GetApplicationChanges(ctx context.Context, eventInfo webhook.EventInfo) ([]
...
+// nested app-of-apps children queued only during diffing, but those are only ever queued by
+// an already-matching parent app, which this function would already have found.
+func HasMatchingApplications(ctx context.Context, eventInfo webhook.EventInfo) (bool, error) {
+	argoApps, err := listApplications(ctx)
+	if err != nil {
+		return false, err
+	}
+	if len(argoApps.Items) == 0 {
+		return false, fmt.Errorf("empty ArgoCD app list")
+	}
```
vrivellino, line 405: bug: an empty app list is an error here, so the flag posts a red `error` commit status on every PR in every repo the webhook sees — the exact noise the flag exists to remove, and the opposite of the README's "stay completely silent (no commit status, no comment) when nothing matches". `GetApplicationChanges` needs this guard because an empty list means it can never produce a result; a "does anything match" query can just answer no.

```suggestion
	if len(argoApps.Items) == 0 {
		log.Warn().Msg("empty ArgoCD app list; treating event as unmatched")
		return false, nil
	}
```

internal/argocd/helper.go
```diff
@@ -388,6 +388,35 @@ func GetApplicationChanges(ctx context.Context, eventInfo webhook.EventInfo) ([]
...
 }
 
+// HasMatchingApplications reports whether any ArgoCD application matches eventInfo, without
+// diffing anything. It reuses the same listing + filtering (including the
+// manifest-generate-paths path filter) that GetApplicationChanges uses for its wave 1
+// (single-source) and wave 3 (multi-source) matching, so a false result here guarantees
+// GetApplicationChanges would return an empty appResList and no error. It does not discover
+// nested app-of-apps children queued only during diffing, but those are only ever queued by
+// an already-matching parent app, which this function would already have found.
+func HasMatchingApplications(ctx context.Context, eventInfo webhook.EventInfo) (bool, error) {
```
vrivellino, line 398: nit: this re-implements the front half of `GetApplicationChanges` (list, filter single-source, filter multi-source), and the two can drift silently — add a fourth matching wave to `GetApplicationChanges` and this function keeps returning `false`, so matching PRs get no status and no comment with nothing in the log to explain it. Extracting the list plus both `filterApplications` calls into one helper that both entry points call removes the drift by construction.

internal/process_event/code_change.go
```diff
@@ -129,6 +129,21 @@ func ProcessCodeChange(eventInfo webhook.EventInfo, devMode bool, wg *sync.WaitG
 		eventInfo.ChangedFiles = changedFiles
 	}
 
+	// ARGO_DIFF_SKIP_UNMATCHED_EVENTS: check for a matching ArgoCD application before posting commit status
+	if strings.ToLower(strings.TrimSpace(os.Getenv("ARGO_DIFF_SKIP_UNMATCHED_EVENTS"))) == "true" {
```
vrivellino, line 133: nit: this file's convention for env-driven config is a small pure helper — `processTimeout()`, `reportReserve()` — precisely so `code_change_test.go` can cover it, and `ProcessCodeChange()` itself has no test. A helper would make the parsing (trimming, case-folding, anything but `true`) testable. Named below for the `ARGO_DIFF_REQUIRE_APP_MATCH` rename discussed in the PR thread.

```suggestion
	if requireAppMatch() {
```

internal/process_event/code_change.go
```diff
@@ -129,6 +129,21 @@ func ProcessCodeChange(eventInfo webhook.EventInfo, devMode bool, wg *sync.WaitG
...
+		matched, err := argocd.HasMatchingApplications(ctx, eventInfo)
+		if err != nil {
+			log.Error().Err(err).Msg("argocd.HasMatchingApplications() failed")
+			_ = github.Status(ctx, github.StatusError, err.Error(), eventInfo.RepoOwner, eventInfo.RepoName, eventInfo.Sha, devMode)
+			*callerErr = err
+			return
+		}
+		if !matched {
+			log.Debug().Msgf("Skipping commit status; No ArgoCD application matches %s/%s#%d", eventInfo.RepoOwner, eventInfo.RepoName, eventInfo.PrNum)
```
vrivellino, line 142: nit: `Debug` hides the one line that explains a dropped event at the default log level. An operator asking "why did argo-diff never touch my PR?" sees nothing at all unless they raise `LOG_LEVEL` and reproduce; `checkSource()` already logs its `ARGO_DIFF_SKIP_REF_CHECK` bypass at `Info` for the same reason.

```suggestion
			log.Info().Msgf("Skipping commit status; No ArgoCD application matches %s/%s#%d", eventInfo.RepoOwner, eventInfo.RepoName, eventInfo.PrNum)
```

internal/process_event/code_change.go
```diff
@@ -129,6 +129,21 @@ func ProcessCodeChange(eventInfo webhook.EventInfo, devMode bool, wg *sync.WaitG
...
+	// ARGO_DIFF_SKIP_UNMATCHED_EVENTS: check for a matching ArgoCD application before posting commit status
+	if strings.ToLower(strings.TrimSpace(os.Getenv("ARGO_DIFF_SKIP_UNMATCHED_EVENTS"))) == "true" {
+		matched, err := argocd.HasMatchingApplications(ctx, eventInfo)
+		if err != nil {
+			log.Error().Err(err).Msg("argocd.HasMatchingApplications() failed")
+			_ = github.Status(ctx, github.StatusError, err.Error(), eventInfo.RepoOwner, eventInfo.RepoName, eventInfo.Sha, devMode)
+			*callerErr = err
+			return
+		}
+		if !matched {
```
vrivellino, line 141: nit: an `argo diff` PR comment reaches this same path (`IsRefreshComment()` in `internal/webhook/process.go` sets `Refresh`), so under this flag a human who explicitly asks for a diff on an unmatched repo gets total silence — no comment, no status, no reaction. Worth exempting `eventInfo.Refresh` from the skip, so an explicitly requested run always answers.

internal/argocd/helper.go
TheRealNoob: Good catch on the noise — but I wonder if an empty list is more likely to signal a misconfiguration (wrong ArgoCD permissions, wrong server address, etc.) than a genuinely empty ArgoCD. In that case silently returning `false` would mask a real problem: every PR in every repo would be skipped with no GitHub signal at all, and the only trace would be a `Warn` line in the logs. The current error behavior at least gives operators a visible red status to investigate. Happy to go with your suggestion if you disagree, but wanted to raise it before changing it.

vrivellino: I think the appropriate place to surface misconfiguration errors is via `ConnectivityCheck()` at startup.

By surfacing them per-event means a red status on every PR in every repo the webhook sees. Also worth noting: a wrong server address or bad token already exits non-zero and lands in the `return false, err` branch above, so an empty list with exit 0 is really RBAC or project scoping.

internal/process_event/code_change.go
TheRealNoob: Fixed — error path now uses `context.WithTimeout(context.Background(), reportReserve(timeout))`, consistent with the other reporting calls in ProcessCodeChange().

internal/argocd/helper.go
TheRealNoob: Agreed — fixed. Empty list now returns `false, nil` with a Warn log.

internal/argocd/helper.go
TheRealNoob: Fixed — extracted `listAndMatchApplications()` as a shared helper. Both `GetApplicationChanges` and `HasMatchingApplications` delegate to it now.

internal/process_event/code_change.go
TheRealNoob: Fixed — added `requireAppMatch()` helper with a unit test, matching the `processTimeout()`/`reportReserve()` convention.

internal/process_event/code_change.go
TheRealNoob: Fixed — changed to `log.Info()`.

internal/process_event/code_change.go
TheRealNoob: Fixed — `eventInfo.Refresh` now bypasses the match check so an explicit `argo diff` comment always gets an answer.

TheRealNoob:
Thank you for the review.  You had a couple really good points.  I believe I've addressed all points so ready for re-review.

internal/process_event/code_change.go
```diff
@@ -130,6 +137,26 @@ func ProcessCodeChange(eventInfo webhook.EventInfo, devMode bool, wg *sync.WaitG
 		eventInfo.ChangedFiles = changedFiles
 	}
 
+	// ARGO_DIFF_REQUIRE_APP_MATCH: check for a matching ArgoCD application before posting
+	// commit status, and stay silent when nothing matches. Explicit refresh requests
+	// (e.g. "argo diff" PR comments) bypass this so a human asking for a diff always gets
+	// an answer.
+	if requireAppMatch() && !eventInfo.Refresh {
```
vrivellino, line 144: [bug] `eventInfo.Refresh` is set unconditionally in GitHub Actions mode  (`internal/server/run_once.go:110`), so this condition is never true under Actions and the flag does nothing there - contradicting the README row added in this PR ("GitHub Actions mode ... honors this flag too").

`Refresh` carries two meanings today, "re-read PR metadata from the API" and "a human asked for this run", and this bypass wants only the second; splitting them into separate `EventInfo` fields fixes it, otherwise the README and `context.md` should say the flag is inert under Actions.

internal/process_event/context.md
```diff
@@ -62,6 +67,10 @@ into an exit code.
   since that invariant lives in another package.
 - No changes, no warnings, and nothing skipped → `github.Comment()` is called with an **empty**
   body list, which clears out any stale argo-diff comments.
+- `unknownCount` is vestigial: it is declared and reported but never incremented.
```
vrivellino: [nit] `unknownCount` no longer exists in `code_change.go` - #314 removed it, and this line is absent  from `main`, so the merge in 3dfeb8b resurrected it. Drop the line

vrivellino:
@TheRealNoob: There's one minor issue, plus a few small merge conflicts to resolve following #311's merge

vrivellino:
@TheRealNoob: FYI - I just published [v3.0.0-rc1](https://github.com/vince-riv/argo-diff/releases/tag/3.0.0-rc1) (and [chart](https://github.com/vince-riv/argo-diff/releases/tag/chart-3.0.0-rc1)). These include your recent contributions. I'll publish `rc2` once this PR lands.

I haven't done an extensive burn-in yet, but I updated the end-to-end workflow to include happy path testing of app-any-namespace.

internal/process_event/code_change.go
TheRealNoob: while this is true, however action.yaml doesn't define `ARGO_DIFF_REQUIRE_APP_MATCH` meaning `requireAppMatch()` will always resolve to false, unless the user defines the env var themselves.  so this should be a hypothetical failure scenario.  You do have a valid point about the double-meaning though so if you'd still like it changed I can do that.

internal/process_event/context.md
TheRealNoob: Fixed — removed the stale line.

TheRealNoob:
Thank you for publishing the RCs :) I will definitely make use of it once this feature lands.  Really appreciate you all doing this project.  I anticipate being a happy consumer so I'm more than happy to give back.

internal/process_event/code_change.go
vrivellino: Ok - I think the code is in a good place as is. We probably just need a few tweaks to docs around this - I'll take care of that in a separate follow-up.

vrivellino (approved):

vrivellino:
> Thank you for publishing the RCs :)

[3.0.0-rc2](https://github.com/vince-riv/argo-diff/releases/tag/3.0.0-rc2) and [chart](https://github.com/vince-riv/argo-diff/releases/tag/chart-3.0.0-rc2) are published!

PR #319, opened 2026-09-10
feat: make comment auto-collapse thresholds configurable

brushmtn:
Makes the "auto" collapse thresholds (currently hardcoded at 3 applications / 5 resources) configurable via two new env vars, `ARGO_DIFF_COMMENT_COLLAPSE_APP_COUNT` and `ARGO_DIFF_COMMENT_COLLAPSE_RESOURCE_COUNT`, and flips the default `ARGO_DIFF_COMMENT_COLLAPSE` mode from `auto` to `expanded`. The auto-collapse feature (#314) has not shipped in a release yet, so this changes the shipped default rather than breaking a released behavior.


brushmtn:

<details>
<summary>Plan</summary>

1. `internal/github/markdown.go`: rename constants to `defaultCollapseAppCount`/`defaultCollapseResourceCount`, add `collapseAppCount()`/`collapseResourceCount()` env-backed getters with non-positive fallback, flip `collapseMode()`'s `""` case to `collapseExpanded`, point `appOpen()`/`resourceOpen()` at the new getters.
2. `internal/github/markdown_test.go`: update `TestCollapseMode` for the new default and add threshold/fallback rows; fix "app carrying a notice" test to explicitly set `auto` mode.
3. Config surfaces: README.md env var table, action.yml inputs, charts/argo-diff/values.yaml + configmap template + unittest, docs/k8s/deployment.yaml, internal/server/run_once.go nonSensitiveVars.
4. Context files: internal/github/context.md, charts/context.md.
5. Verify: go test ./..., golangci-lint run, helm unittest, helm-docs.

</details>

<details>

**Done — plan complete, plus review fixes:**

All original plan steps landed (see prior ledger entries in this comment's history / commits f0ffc20, 2bcb846). vrivellino's review on commit 2bcb846 flagged two issues, both fixed in commit 9174f9e:

1. `collapseMode()`'s unknown-value fallback still returned `collapseAuto` instead of the new `collapseExpanded` default - fixed in `internal/github/markdown.go`. Updated `TestCollapseMode`'s "nonsense" row to use 9 apps/9 resources (was 1/1) so it actually distinguishes an expanded-fallback result from an auto one, instead of passing either way.
2. Every `ARGO_DIFF_COMMENT_COLLAPSE_RESOURCE_COUNT` description incorrectly said it folds "an application's block" - that's `ARGO_DIFF_COMMENT_COLLAPSE_APP_COUNT`'s job (`appOpen()`). The resource-count threshold folds an application's individual resource (diff) blocks (`resourceOpen()`). Fixed in README.md, action.yml, charts/argo-diff/values.yaml (+ regenerated charts/argo-diff/README.md via helm-docs), and internal/github/context.md.

The review also noted commit 2bcb846 used a non-standard `docs+chart:` subject type (repo convention is `feat:`/`fix:`/`chore:`/`docs:`), but flagged it as non-blocking since squash-merge uses the PR title, not individual commit subjects - left as-is rather than rewriting pushed history.

**Verified after fixes:** `go build ./...`, `go test ./...` (including the corrected TestCollapseMode case), `golangci-lint run` (0 issues), `gofmt -l .` (clean), `helm unittest charts/argo-diff` (61/61 passed).

**No blockers.** Ready for re-review.

</details>

internal/github/markdown.go
```diff
@@ -126,11 +127,35 @@ func indexCount() int {
...
+	}
+	return n
+}
+
 func collapseMode() string {
 	raw := strings.ToLower(strings.TrimSpace(os.Getenv("ARGO_DIFF_COMMENT_COLLAPSE")))
 	switch raw {
 	case "":
-		return collapseAuto
+		return collapseExpanded
```
vrivellino, line 158: bug: the empty default is now `expanded`, but an unrecognized value still falls through to `collapseAuto` below, so `ARGO_DIFF_COMMENT_COLLAPSE=expand` (a typo) silently auto-folds a 10-application comment while the README documents `expanded`. Return `collapseExpanded` from that fallback, and fix the warning text, which still says `using auto`.

README.md
```diff
@@ -249,7 +249,9 @@ the accepted environment variables and their respective GitHub Actions inputs.
...
+| ARGO_DIFF_COMMENT_COLLAPSE       | comment_collapse            | no               | `expanded` | Whether the collapsible sections in a PR comment start open. `expanded` (default) always starts them open; `collapsed` always starts them folded; `auto` keeps a small comment fully expanded and folds one that covers more than `ARGO_DIFF_COMMENT_COLLAPSE_APP_COUNT` applications, or an application with more than `ARGO_DIFF_COMMENT_COLLAPSE_RESOURCE_COUNT` changed resources. |
+| ARGO_DIFF_COMMENT_COLLAPSE_APP_COUNT | comment_collapse_app_count | no           | `3`      | In `auto` collapse mode, the application count above which a comment is folded. Must be positive; a non-positive value warns and falls back to the default. |
+| ARGO_DIFF_COMMENT_COLLAPSE_RESOURCE_COUNT | comment_collapse_resource_count | no | `5`      | In `auto` collapse mode, the changed-resource count above which an application's block is folded. Must be positive; a non-positive value warns and falls back to the default. |
```
vrivellino: bug: this threshold folds the resource blocks inside an application, not the application's own block — `resourceOpen()` drives only the resource `<details>`, and `TestCollapseMode`'s 1-app/9-resource row asserts the app block stays `<details open>`. The same wording needs fixing in `action.yml`, `charts/argo-diff/values.yaml` and `charts/argo-diff/README.md`.

```suggestion
| ARGO_DIFF_COMMENT_COLLAPSE_RESOURCE_COUNT | comment_collapse_resource_count | no | `5`      | In `auto` collapse mode, the changed-resource count above which that application's resource blocks are folded. Must be positive; a non-positive value warns and falls back to the default. |
```

internal/github/markdown_test.go
```diff
@@ -274,21 +275,30 @@ func TestIndexCount(t *testing.T) {
...
-		{"nonsense", 1, 1, "<details open>", "<details open>"}, // falls back to auto
+		{"", "", "", 1, 1, "<details open>", "<details open>"},         // default mode is expanded, whatever the counts
+		{"auto", "", "", 1, 1, "<details open>", "<details open>"},     // auto, small
+		{"auto", "", "", 9, 1, "<details>", "<details open>"},          // auto, many apps
+		{"auto", "", "", 1, 9, "<details open>", "<details>"},          // auto, many resources
+		{"auto", "10", "", 9, 1, "<details open>", "<details open>"},   // raised app threshold keeps it open
+		{"auto", "", "10", 1, 9, "<details open>", "<details open>"},   // raised resource threshold keeps it open
+		{"expanded", "", "", 9, 9, "<details open>", "<details open>"}, // forced open
+		{"collapsed", "", "", 1, 1, "<details>", "<details>"},          // forced closed
+		{"nonsense", "", "", 1, 1, "<details open>", "<details open>"}, // falls back to auto
```
vrivellino: nit: at 1 application and 1 resource both `auto` and `expanded` render open, so this row passes under either fallback and cannot catch a change to it. Raise it to 9 applications to pin the fallback down.

vrivellino (commented):
Two fixes before merge: `collapseMode()`'s unknown-value fallback still returns `auto` instead of the new `expanded` default, and the new `..._RESOURCE_COUNT` docs name the wrong element as the one that folds. Every config surface is covered and all CI checks pass.

- Commit `2bcb846` uses the subject type `docs+chart:`; `CLAUDE.md` lists `feat:`, `fix:`, `chore:` and `docs:`. Squash-merging on the PR title keeps it off `main`.

vrivellino (commented):
Re-reviewed at `9174f9e`. All three findings are resolved and I found nothing new: the unknown-value fallback and its warning text now both say `expanded`, the `nonsense` test row at 9 apps/9 resources genuinely distinguishes the `expanded` fallback from `auto`, and the `..._RESOURCE_COUNT` wording is corrected in `README.md`, `action.yml`, `values.yaml`, the regenerated chart README and `internal/github/context.md`. `action.yml` still parses, and every check passes.

- Still outstanding, non-blocking: commit `2bcb846` uses the subject type `docs+chart:`, which is not one of the types `CLAUDE.md` lists. Squash-merging on the PR title keeps it off `main`.

vrivellino (approved):

Issue #317, opened 2026-09-10
chart: add top-level env and envFrom values for arbitrary container environment


vrivellino:
## Problem

Every argo-diff environment variable that an operator wants to set through the chart needs a bespoke `config.*` values key. There is no escape hatch: `templates/deployment.yaml` hardcodes the container's `env:` (just `LOG_LEVEL`, plus `ARGOCD_CLI_CMD_NAME` when `argocdCli.image.tag` is set) and its `envFrom:` (exactly the chart's own ConfigMap and Secret). `templates/configmap.yaml` builds that ConfigMap from a fixed allowlist.

`charts/context.md` already records the consequence:

> **Not every argo-diff env var has a values key** — `ARGO_DIFF_TIMEOUT` and `ARGO_DIFF_DISABLE_NON_GITHUB_REPO_MATCH`, for example, currently require an externally managed ConfigMap. Adding one means touching `values.yaml` (with its `# key -- description` comment for helm-docs), `templates/configmap.yaml`, `tests/configmap_test.yaml`, and the README table.

So an operator has two bad options today: fork the chart, or set `config.configMapCreate: false` and hand-manage the whole ConfigMap — which also gives up every typed `config.*` key at once.

This surfaced in #313, where adding one new variable cost five files of chart plumbing.

## Proposal

Add two new top-level values that pass straight through to the Deployment's container spec.

### `env`

A raw list of core/v1 `EnvVar` entries, appended after the chart's own entries.

```yaml
# templates/deployment.yaml
          env:
            - name: LOG_LEVEL
              value: "{{ .Values.logLevel }}"
            {{- if .Values.argocdCli.image.tag }}
            - name: ARGOCD_CLI_CMD_NAME
              value: "/argocd-cli/argocd"
            {{- end }}
            {{- with .Values.env }}
            {{- toYaml . | nindent 12 }}
            {{- end }}
```

Raw passthrough (not a `name: value` map) so `valueFrom` keeps working — `secretKeyRef`, `configMapKeyRef`, `fieldRef`.

```yaml
# values.yaml
env: []
#  - name: ARGO_DIFF_TIMEOUT
#    value: "5m"
#  - name: ARGO_DIFF_REFRESH_COMMENT_KEYWORDS
#    value: "argo diff,argo-diff,please diff"
#  - name: GITHUB_TOKEN
#    valueFrom:
#      secretKeyRef:
#        name: my-gh-token
#        key: token
```

### `envFrom`

[...]

PR #320, opened 2026-09-11
feat(chart): add deployment.env and deployment.envFrom passthrough

brushmtn:
Adds `deployment.env` and `deployment.envFrom` to the `argo-diff` Helm chart as a raw passthrough for the Deployment's container `env`/`envFrom`, closing #317. Not every argo-diff environment variable has a typed `config.*` values key, and adding one touches five files (`values.yaml`, `templates/configmap.yaml`, `tests/configmap_test.yaml`, the README table); this is an escape hatch for the rest, appended after the chart's own entries so an operator-supplied `env` name or `envFrom` source always wins on a collision. It also lets an operator layer extra ConfigMaps/Secrets on top of the chart-managed one, which `config.configMapCreate: false` cannot do without giving up every typed `config.*` key.

Values were placed under `deployment.env` / `deployment.envFrom` (resolving the issue's open question) rather than top-level, matching the existing `deployment.volumes` / `deployment.volumeMounts` pattern for other raw pod-spec passthrough knobs in this chart.

Updated: `templates/deployment.yaml` (the two `with`/`toYaml` blocks), `values.yaml` (new keys with helm-docs comments and commented examples), `tests/deployment_test.yaml` (defaults unchanged, append order, `valueFrom` survives passthrough, duplicate-name override of `LOG_LEVEL` and `ARGOCD_CLI_CMD_NAME`, `envFrom` append order), regenerated `charts/argo-diff/README.md` via `helm-docs`, `charts/context.md`, and the top-level `README.md` env var table intro.

`helm lint` and `helm unittest .` (66 tests) pass.


brushmtn:

<details>
<summary>Plan</summary>

1. Read issue #317 and decide the open question: `deployment.env`/`deployment.envFrom` (matches `deployment.volumes`/`deployment.volumeMounts` convention) vs top-level `env`/`envFrom`. Chose `deployment.*`.
2. Add the two `with`/`toYaml` passthrough blocks to `templates/deployment.yaml`, appended after the chart's own `env`/`envFrom` entries.
3. Add `deployment.env: []` / `deployment.envFrom: []` to `values.yaml` with helm-docs `# key -- description` comments and commented examples, documenting the `checksum/config`/`checksum/secret` gap for `envFrom`.
4. Add unittest coverage in `tests/deployment_test.yaml`: defaults unchanged, append order, `valueFrom` survives passthrough, duplicate name overrides `LOG_LEVEL` and `ARGOCD_CLI_CMD_NAME` by landing after them, `envFrom` refs land after the chart's own.
5. Regenerate `charts/argo-diff/README.md` via `helm-docs --chart-search-root=charts`.
6. Update `charts/context.md`: replace the "Not every argo-diff env var has a values key" paragraph, document the new escape hatch and the checksum gap.
7. Update top-level `README.md`'s env var table intro to mention `deployment.env`/`deployment.envFrom`.
8. Run `helm lint` and `helm unittest .`; verify all pass.
9. Commit, push, open PR.

</details>

<details>

Done: all plan steps complete, plus review fixes. vrivellino's review on commit d913ab5 flagged one real bug and two doc-wording nits, all fixed in commit ea50ca0:

1. Bug: setting `deployment.env` with a name matching `LOG_LEVEL` or `ARGOCD_CLI_CMD_NAME` rendered two `env` entries sharing that name, which Kubernetes rejects under server-side apply (`duplicate entries for key [name="LOG_LEVEL"]`). Fixed `templates/deployment.yaml` to compute the set of names in `deployment.env` up front and omit the chart's own entry for a name already present there, instead of appending both. Verified with `helm template` that overriding either name now renders exactly one entry. Updated the two duplicate-name unittest cases in `tests/deployment_test.yaml` to assert single-entry output.
2. Nit: the `deployment.envFrom` values comment implied it could win over `LOG_LEVEL`/`ARGOCD_CLI_CMD_NAME`, but Kubernetes always resolves every `envFrom` source before the container's own `env:` entries, so it never can — only `deployment.env` can override those. Reworded in `values.yaml` and `charts/context.md`.
3. Nit: the top-level `README.md` env var table intro claimed every variable in the table has a typed `config.*` key, which the next sentence then contradicted, and not every typed key lives under `config.*` anyway (`logLevel`, `secret.*`). Applied the reviewer's suggested replacement text verbatim.

Replied to all three review threads on the PR referencing commit ea50ca0.

Verified after fixes: `helm lint charts/argo-diff` passes (only the pre-existing "icon is recommended" info notice), `helm unittest charts/argo-diff` passes all 66 tests across 7 suites, and `helm template` output manually checked for both the default case and the override case (no duplicate `env` names, override value wins).

No blockers. Ready for re-review.

</details>

charts/argo-diff/templates/deployment.yaml
```diff
@@ -80,11 +80,17 @@ spec:
             - name: ARGOCD_CLI_CMD_NAME
               value: "/argocd-cli/argocd"
             {{- end }}
+            {{- with .Values.deployment.env }}
+            {{- toYaml . | nindent 12 }}
```
vrivellino, line 90: bug: the documented duplicate-name override renders two `env` entries with the same `name`, and `env` is an associative list keyed by `name`, so server-side apply rejects the object — Argo CD with `ServerSideApply=true` (or `helm upgrade --server-side`) fails with `duplicate entries for key [name="LOG_LEVEL"]`. Emit the chart's `LOG_LEVEL` / `ARGOCD_CLI_CMD_NAME` entries only when `deployment.env` holds no entry of that name, so the rendered list stays unique.

charts/argo-diff/values.yaml
```diff
@@ -201,6 +201,30 @@ deployment:
...
+  #    valueFrom:
+  #      secretKeyRef:
+  #        name: my-gh-token
+  #        key: token
+
+  # deployment.envFrom -- Additional envFrom sources for the argo-diff container, appended after the
+  # chart's own ConfigMap/Secret refs, so these win on key collisions. Raw core/v1 EnvFromSource
+  # entries. Note: the pod's checksum/config and checksum/secret annotations only hash the chart's
+  # own configmap.yaml/secret.yaml, so a change to a ConfigMap/Secret referenced here does not by
+  # itself trigger a rollout.
```
vrivellino: nit: a key supplied through `deployment.envFrom` can never override `LOG_LEVEL` or `ARGOCD_CLI_CMD_NAME`, because the kubelet resolves every `envFrom` source first and then applies `env` over the result. An operator who puts `LOG_LEVEL: debug` in an extra ConfigMap referenced here still gets `info`, so the collision note is worth qualifying.

vrivellino (commented):
One thing to settle before merge: the documented duplicate-name override renders two `env` entries with the same `name`, which server-side apply rejects. The other two findings are documentation wording. Helm lint, 66 unittests, and the k3s e2e check all pass.

README.md
```diff
@@ -235,6 +235,10 @@ the accepted environment variables and their respective GitHub Actions inputs.
 > **Important:** Valid GitHub API credentials are required to run — either the `GITHUB_APP_*` variables
 > must be set, or `GITHUB_TOKEN` / `GITHUB_PERSONAL_ACCESS_TOKEN` must be set.
 
+Any of these variables can be set through the Helm chart via its typed `config.*` values (see
+`charts/argo-diff/README.md`). Any variable without a typed key — including ones not listed here —
+can still be set through the chart's `deployment.env` / `deployment.envFrom` passthrough values.
```
vrivellino: nit: the first sentence says every variable in this table is settable via a typed `config.*` value, which the second sentence then contradicts — and the typed keys are not all under `config.*` (`LOG_LEVEL` is `logLevel`, `ARGOCD_AUTH_TOKEN` / `GITHUB_TOKEN` live under `secret.*`).

```suggestion
Many of these variables have a typed Helm values key (`config.*`, `logLevel`, `secret.*`) — see
`charts/argo-diff/README.md`. Any variable without a typed key — including ones not listed here —
can still be set through the chart's `deployment.env` / `deployment.envFrom` passthrough values.
```

charts/argo-diff/templates/deployment.yaml
brushmtn: Good catch. Fixed in ea50ca0: templates/deployment.yaml now checks deployment.env for the name first and omits the chart's own LOG_LEVEL/ARGOCD_CLI_CMD_NAME entry when a same-named override is present, instead of rendering both. Verified with `helm template` that no duplicate names render, and updated the two duplicate-name unittest cases to assert single-entry output.

charts/argo-diff/values.yaml
brushmtn: Fixed in ea50ca0: reworded the deployment.envFrom comment to make clear it only wins on key collisions against the chart's own ConfigMap/Secret refs, and can never override LOG_LEVEL/ARGOCD_CLI_CMD_NAME since Kubernetes always resolves envFrom before the container's env: entries.

README.md
brushmtn: Applied your suggestion verbatim in ea50ca0.

charts/argo-diff/README.md
```diff
@@ -134,3 +136,4 @@ $ helm install my-release oci://ghcr.io/vince-riv/chart/argo-diff
 
 ----------------------------------------------
 Autogenerated from chart metadata using [helm-docs v1.14.2](https://github.com/norwoodj/helm-docs/releases/v1.14.2)
+
```
vrivellino, line 139: nit: this trailing blank line is not helm-docs output — `main` ends the file right after the footer line. The next `helm-docs` run strips it, so it shows up as unrelated noise in a later chart diff.

vrivellino (commented):
Re-reviewed at `ea50ca0e`. All three findings from the previous review are addressed: the template now omits the chart's own `LOG_LEVEL` / `ARGOCD_CLI_CMD_NAME` entry when `deployment.env` supplies that name, and both doc corrections landed. Verified locally against the head commit — `helm template` renders a single entry per name for the override cases, and all 66 unittests pass. One trivial nit left, nothing blocking.

vrivellino (approved):

PR #321, opened 2026-09-13
fix(chart): merge deployment.env with chart defaults by name instead of per-name guards

TheRealNoob:
## Summary

Follow-up to https://github.com/vince-riv/argo-diff/pull/320
Related to https://github.com/vince-riv/argo-diff/issues/317

Reviewing that PR's `deployment.env` passthrough, the duplicate-name handling didn't sit right: it hardcodes a per-name `if not (has "LOG_LEVEL" $envNames)` / `if not (has "ARGOCD_CLI_CMD_NAME" $envNames)` guard for each chart default, which doesn't scale (every future default needs its own guard) and isn't actually doing the dedup itself — `append` only builds the `$envNames` lookup list the guards check against.

This replaces that with a single `argo-diff.env` helper (`templates/_helpers.tpl`) that merges `deployment.env` over the chart's defaults by name in one Sprig `merge` call, so a same-named entry always replaces the default instead of needing a dedicated guard.

One subtlety worth flagging for review: the two chart defaults are kept as plain scalars rather than `{value: ...}` maps. This is deliberate — Sprig's `merge`/`mergeOverwrite` deep-merge nested maps (via mergo), so if the defaults were also `{value: ...}` maps, an override that switches the same name to `valueFrom` would have the default's `value` field survive alongside it (Kubernetes rejects an `EnvVar` with both set). Keeping defaults as scalars means a type mismatch prevents that recursion. Covered by a dedicated test.

This also fixes a coercion issue the merge otherwise introduces: a non-string `logLevel` (eg. `--set logLevel=5`) previously rendered as an unquoted YAML number via `toYaml`, whereas the old template's literal `value: "{{ .Values.logLevel }}"` always quoted it. Now explicitly cast to a string.

One user-visible behavior change: the rendered `env:` list is now sorted alphabetically by name rather than defaults-then-appended-entries. Noted in `charts/context.md`.

`helm lint` and `helm unittest .` (66 tests) pass.

## Test plan

- [x] `helm lint charts/argo-diff`
- [x] `helm unittest charts/argo-diff` (66/66 passing)
- [x] Manually rendered `deployment.yaml` with `--set logLevel=5`, `deployment.env` overrides, and `argocdCli.image.tag` set to confirm ordering, overrides, and the value/valueFrom leak scenario

Assisted-by: Claude Code (claude-opus-4-6)


TheRealNoob:
This won't affect my use cases so it won't bother me if you don't merge it.  I was just reading over the last PR and couldn't understand how it solved some of the stated problems and felt I had to at least share how I've seen this solved in other charts.

vrivellino (approved):
Thanks @TheRealNoob!

PR #324, opened 2026-09-13
feat(k3s): overridable bundled argocd CLI version + app-in-any-namespace e2e coverage

brushmtn:
This adds two independent capabilities to the k3s e2e test in `.github/workflows/k3s.yml`: a `workflow_dispatch` input to override the version of the `argocd` CLI bundled into the argo-diff image under test (via a new `ARGOCD_IMAGE` Dockerfile build-arg, which Renovate already tracks automatically), and a second, parallel app-of-apps tree (`any-ns-meta`/`any-ns-deployment`, split across the `argocd-apps` and `apps` namespaces) that exercises argo-diff's `--app-namespace` support for ArgoCD's app-in-any-namespace feature, added in #311. The existing `argocd`-namespace apps are untouched; both scenarios now run side by side. The in-cluster ArgoCD chart/app version and the runner's own `argocd` CLI (used to drive `argocd app wait`) still float on "latest", unchanged — pinning those is a possible future enhancement, noted in `test/context.md` but not built here.

A pinned `argocd_version` dispatch now also gets its own PR comment: `ARGO_DIFF_CONTEXT_STR` (the key argo-diff uses to find its own comment on a PR) varies with the version, and `ARGO_DIFF_COMMENT_PREAMBLE` calls out which version is under test, so a version-override run no longer overwrites the standard run's comment on the same PR.

**Depends on #311.** `--app-namespace` support doesn't exist on `main` yet (#311 is an open, unmerged, cross-repo fork PR), so `argo-diff`'s `app diff` calls against `any-ns-meta`/`any-ns-deployment` currently fail CI with a `PermissionDenied` from ArgoCD — expected until #311 merges, not a bug in this PR. Everything up to that point (Dockerfile override, namespace/RBAC/AppProject wiring, ArgoCD sync of the any-ns apps) is verified working on real CI. Once #311 lands on `main`, rebase this branch and CI should go green.


brushmtn:

<details>
<summary>Plan</summary>

1. Dockerfile: replace the fixed `FROM quay.io/argoproj/argocd:...` with an `ARG ARGOCD_IMAGE=quay.io/argoproj/argocd:v3.5.2@sha256:...` default, `FROM ${ARGOCD_IMAGE} AS argocd`, and a `RUN argocd version --client` build-time sanity check. No renovate.json change needed — Renovate's Docker manager already tracks this `ARG X=image:tag@digest` + `FROM ${X}` pattern.

2. `.github/workflows/k3s.yml`: add an optional `workflow_dispatch.inputs.argocd_version` (default `''`). Split the existing "Build argo-diff image" step into two, gated by `if: ${{ github.event.inputs.argocd_version == '' }}` (unchanged default build) vs `!= ''` (new step passing `--build-arg ARGOCD_IMAGE=quay.io/argoproj/argocd:<version>`). The in-cluster ArgoCD install and the runner's own `argocd` CLI install stay unchanged (latest).

3. Add a second, parallel app-of-apps tree for app-in-any-namespace coverage, additive to the existing `argocd`-namespace apps: `test/argocd-meta-any-ns.yaml` (AppProject `any-ns` + Application `any-ns-meta` in namespace `argocd-apps`), `test/argocd-applications-any-ns/` (child Application `any-ns-deployment`, kustomized into namespace `apps`), `test/any-ns-deployment/` (the actual workload namespace + Deployment, self-contained so it can't collide with `basic-deployment`'s cluster-scoped resources).

4. `test/argocd-helmchart.yaml`: add `configs.params."application.namespaces": "argocd-apps,apps"` so ArgoCD watches Applications outside its own namespace.

5. `.github/workflows/k3s.yml`: create the `argocd-apps`/`apps` namespaces, apply `argocd-meta-any-ns.yaml`, and add `argocd app wait any-ns-meta --app-namespace argocd-apps` / `argocd app wait any-ns-deployment --app-namespace apps` to the existing wait step. The broken-app/failure-scenario steps stay `argocd`-namespace-only, unchanged.

6. Update `test/context.md` with the new namespaces, the `argocd_version` dispatch input, and a note that pinning the in-cluster ArgoCD version is a possible future enhancement, not covered here.

7. Verify: `go build -v ./...` and `go test ./...` (no Go changes, should stay green), YAML-parse-check all new/changed YAML files (no Docker available in this sandboxed environment to build-test the Dockerfile locally — that needs CI or a manual dispatch). Commit, push, open PR.

</details>

<details>

Status: implementation complete; CI is red but for an expected, documented reason (blocked on #311), not a bug in this branch. Do not spend more CI cycles debugging the "Run argo-diff (healthy)" failure below — it will resolve itself once #311 merges to main and this branch rebases.

Done: all plan steps implemented and pushed (commits 0bfcd11 through b39105d), plus one additive commit to the shared `k3s-test` branch (c883a19, "chore(gha): add any-ns app-in-any-namespace e2e fixtures") adding the new fixture paths there too — required because the any-ns Applications pin `targetRevision: k3s-test`, same as `meta`/`basic-deployment`, and that stable branch didn't have the new paths yet. User approved that push explicitly (it's a shared branch other PRs' e2e runs also depend on) before it happened.

Two real bugs were found via live CI runs and fixed on this branch: (1) the `ARGOCD_IMAGE` build-arg was declared after the first `FROM`, scoping it to the wrong stage so it resolved blank in the second `FROM` — fixed by moving the `ARG` to global scope, before any `FROM` (commit 47bad5a). (2) `kubectl -n argocd apply -f test/argocd-meta-any-ns.yaml` conflicted with that file's own `metadata.namespace: argocd-apps` on the Application object — fixed by giving the AppProject its own explicit `namespace: argocd` and dropping the `-n` flag for that one apply (commit 9e91058). A third change, setting `configs.rbac.policy.default: role:admin` in `test/argocd-helmchart.yaml` (commit b39105d), fixes a real, independent, documented upstream RBAC gap (argoproj/argo-cd#332: `server.disable.auth` bypasses authentication but not RBAC, and the built-in `role:readonly` default policy doesn't cover the `<project>/<namespace>/<app>` object format apps outside the `argocd` namespace use) — keep this even though it wasn't sufficient on its own to fix the current CI failure, since it's still a correctness fix that will matter once #311 lands.

Verified via live CI (not just locally): Docker build now succeeds with the ARGOCD_IMAGE fix. The any-ns AppProject, both any-ns Applications, and their RBAC/namespace wiring all sync and reach Healthy/Synced correctly (confirmed via a dedicated debug step added at commit db555ec, and via a clean "Sync Application and wait for ready" pass in the most recent run) — this proves the fixtures, the `application.namespaces` ArgoCD config, and the k3s-test branch update are all correct.

Root cause of the remaining CI failure: `internal/argocd` on `main` (and therefore on this branch, which was correctly branched from a fresh `origin/main` per this repo's workflow) has no `--app-namespace` flag support at all — that code lives entirely in #311, an open, unmerged, cross-repo fork PR (`TheRealNoob:fix/app-namespace-for-multi-namespace-apps`, not authored by this session). Without that flag, `argocd app diff any-ns-meta`/`any-ns-deployment` (run without `--app-namespace`) gets a gRPC `PermissionDenied` from the ArgoCD server. This is the expected, by-design failure mode until #311 merges — the whole point of this PR's e2e coverage is to exercise code that doesn't exist on this branch yet. Confirmed by grepping this branch's `internal/argocd/*.go` for `app-namespace`/`AppNamespace`: zero matches.

Additional change requested and made after the above (commit a80665c): a pinned `argocd_version` dispatch now gets its own PR comment instead of overwriting the standard run's. `ARGO_DIFF_CONTEXT_STR` is the key `internal/github/comment.go` uses to find/match an existing comment on the PR (`commentIdentifier`); it was previously hardcoded to `ephemeral-environment-test`/`-failure` regardless of `argocd_version`, so a version-override dispatch would silently overwrite the default run's comment on the same PR. Both "Run argo-diff (healthy)" and "Run argo-diff (manifest failure)" steps now vary `ARGO_DIFF_CONTEXT_STR` on `argocd_version` (appending `-argocd-<version>`) and set `ARGO_DIFF_COMMENT_PREAMBLE` to call out which version is under test (the preamble otherwise silently falls back to the raw context string). `test/context.md` documents this. Not yet verified live (requires a dispatch with a non-blank `argocd_version`, which itself requires #311 first per the blocker above, since the healthy-run diff would still fail before posting a final comment) — logic was verified by reading `internal/github/comment.go`'s `commentIdentifier`/`wrapComment`/`boundPreamble` directly and matches the existing `cond && 'x' || 'y'` ternary style already used elsewhere in this repo's workflows (e.g. `dockerbuild.yml`).

Review feedback (vrivellino) on commit a80665c flagged a real bug in both `ARGO_DIFF_COMMENT_PREAMBLE` lines, confirmed live in this PR's own CI comment: `cond && '' || format(...)` collapses to the `format(...)` branch whenever `cond` is true, because GitHub Actions expressions treat `''` as falsy, so every default (non-override) run was evaluating the override message with a blank version instead of `''`. Fixed in commit bf878f8 by swapping which branch is guarded by `&&` (`cond != '' && format(...) || ''`), exactly per the reviewer's suggested diff. Replied to both review threads referencing bf878f8. Not yet re-verified live for the default (blank `argocd_version`) case specifically, since that requires the "Run argo-diff (healthy)" step to actually post its final comment, which the #311 blocker above prevents — the logic fix itself is confirmed correct by re-reading GitHub Actions' documented truthy/falsy expression rules against the corrected expression.

What's left: nothing further from this session until #311 merges to `main`. Once it does: rebase this branch on the updated `main`, push, and CI should go green (or surface any remaining real issues in #311's own `--app-namespace` wiring, which would then be genuinely new information). At that point also worth doing one dispatch with a non-blank `argocd_version` to confirm the new per-version comment actually posts as a separate comment on the test PR, and confirming the default (blank-version) run's comment no longer carries a stray preamble line. No other blockers.

</details>

.github/workflows/k3s.yml
```diff
@@ -227,13 +252,35 @@ jobs:
...
         timeout-minutes: 2
         env:
           POD_NAME_PREFIX: argo-diff-healthy
           IMAGE: ${{ env.ARGO_DIFF_IMAGE }}
-          ARGO_DIFF_CONTEXT_STR: ephemeral-environment-test
+          # A non-default argocd_version gets its own context string (and thus
+          # its own PR comment) so a pinned-version dispatch doesn't clobber
+          # the standard run's comment on the same PR.
+          ARGO_DIFF_CONTEXT_STR: ${{ github.event.inputs.argocd_version == '' && 'ephemeral-environment-test' || format('ephemeral-environment-test-argocd-{0}', github.event.inputs.argocd_version) }}
+          ARGO_DIFF_COMMENT_PREAMBLE: ${{ github.event.inputs.argocd_version == '' && '' || format('Testing with bundled argocd CLI `{0}` (overridden via workflow_dispatch input `argocd_version`).', github.event.inputs.argocd_version) }}
```
vrivellino: bug: the `cond && '' || format(...)` pattern collapses to the `format(...)` branch whenever `cond` is true and the empty string is falsy, so on every non-override run (`argocd_version == ''`) this evaluates to the override message with an empty version instead of `''`. Confirmed in this PR's own CI run: with `argocd_version` unset, the healthy-run step logs `ARGO_DIFF_COMMENT_PREAMBLE: Testing with bundled argocd CLI `` (overridden via workflow_dispatch input `argocd_version`).` instead of blank.
```suggestion
          ARGO_DIFF_COMMENT_PREAMBLE: ${{ github.event.inputs.argocd_version != '' && format('Testing with bundled argocd CLI `{0}` (overridden via workflow_dispatch input `argocd_version`).', github.event.inputs.argocd_version) || '' }}
```

.github/workflows/k3s.yml
```diff
@@ -262,7 +309,10 @@ jobs:
         env:
           POD_NAME_PREFIX: argo-diff-failure
           IMAGE: ${{ env.ARGO_DIFF_IMAGE }}
-          ARGO_DIFF_CONTEXT_STR: ephemeral-environment-test-failure
+          # See the healthy-run step above: a non-default argocd_version gets
+          # its own context string/comment here too.
+          ARGO_DIFF_CONTEXT_STR: ${{ github.event.inputs.argocd_version == '' && 'ephemeral-environment-test-failure' || format('ephemeral-environment-test-failure-argocd-{0}', github.event.inputs.argocd_version) }}
+          ARGO_DIFF_COMMENT_PREAMBLE: ${{ github.event.inputs.argocd_version == '' && '' || format('Testing with bundled argocd CLI `{0}` (overridden via workflow_dispatch input `argocd_version`) - manifest failure scenario.', github.event.inputs.argocd_version) }}
```
vrivellino: bug: same falsy-collapse issue as the healthy-run preamble above — swap which branch is the guard so the potentially-empty value sits in the `||` fallback, not after `&&`.
```suggestion
          ARGO_DIFF_COMMENT_PREAMBLE: ${{ github.event.inputs.argocd_version != '' && format('Testing with bundled argocd CLI `{0}` (overridden via workflow_dispatch input `argocd_version`) - manifest failure scenario.', github.event.inputs.argocd_version) || '' }}
```

vrivellino (commented):
No changes required before merge, but the two `ARGO_DIFF_COMMENT_PREAMBLE` lines have a real bug (see inline comments) that's already visible in this PR's own CI run — every default (non-override) e2e run posts a stray "Testing with bundled argocd CLI `` (overridden...)" preamble instead of leaving it blank. The `K8s v1.34` check failure itself is expected and already explained in the PR description (depends on unmerged #311), not a defect in this PR.

.github/workflows/k3s.yml
brushmtn: Fixed in bf878f8 — swapped which branch is guarded by `&&` so the potentially-empty value sits in the `||` fallback, matching your suggestion exactly.

.github/workflows/k3s.yml
brushmtn: Fixed in bf878f8 — same swap applied here.

vrivellino (approved):

PR #326, opened 2026-09-14
chore: close test and doc gaps from #311 app-in-any-namespace

brushmtn:
Closes test and doc gaps left by #311 (app-in-any-namespace support): missing coverage for `appKey()`'s namespace disambiguation (the fix for a real wrong-app-diff bug), missing coverage for the version-gate notices, and two stale `context.md` files. Assisted-by: Claude Code (claude-sonnet-5)


brushmtn:

<details>
<summary>Plan</summary>

1. `appKey()` test coverage (highest value — covers a real correctness fix): add `TestAppListToMapKeysByNamespaceAndName` and `TestArgoAppsWithChangesDisambiguatesByNamespace` to `internal/argocd/helper_test.go`; extend `argoAppManifestFixture` with a namespace parameter.
2. Notice test coverage: add two subtests to `TestGetApplicationManifestsAppNamespaceArg` in `internal/argocd/argocd_client_test.go` covering the old-client and lookup-failure notice paths (containment-only assertions, since `resetNotices()` is unexported).
3. Update `internal/config/context.md` and `internal/argocd/context.md` to reflect that `AddNotice()` now has a caller and to document namespace-aware app matching.
4. Verify: `go build`, `go fmt`, `go test -race ./...`, `golangci-lint run`; confirm each new test fails without its fix, then restore.

</details>

<details>

All three gaps from the original brief are complete. Three rounds of review from vrivellino have been addressed.
Round 1 (commit cbd5a9e): fixed the addNotice-seam bug in the two notice subtests, an inverted context.md line about which Applications go missing on an old CLI, and a slices.Concat nit.
Round 2 (commit f5a6ab0): the same namespace-inversion wording bug recurred in four more spots - internal/argocd/context.md's app-in-any-namespace section, the log.Warn() message and the old-client AddNotice() message in argocd_client.go. Also documented the addNotice seam in context.md.
Round 3 (commit eb4c51e): fixed a stale doc comment on supportsManifestsAppNamespace() that said the version check only matters for apps outside the CLI's default namespace; getApplicationManifests() passes --app-namespace for every parent, so an old CLI without the check would reject every argocd app manifests call.
Verification after round 3: go build clean, go fmt no diff, go test -race ./... all pass, golangci-lint run 0 issues.
No blockers. Watching for further review rounds.

</details>

internal/argocd/argocd_client_test.go
```diff
@@ -502,6 +504,53 @@ func TestGetApplicationManifestsAppNamespaceArg(t *testing.T) {
 			t.Errorf("expected --app-namespace to still be applied on the second call, got args %v", captured)
 		}
 	})
+
+	// config.resetNotices() is unexported and only usable inside package
+	// config, so these two subtests can't clear notice state before or after
+	// running. They assert containment only, never absence - AddNotice()
+	// dedupes, so re-running this subtest (or others that raise the same
+	// notice) doesn't change the outcome.
```
vrivellino: bug: In a normal run these containment checks pass without this subtest's own call, because the earlier "client older than 3.5.0" and "can't be determined" subtests already put both notices in the global list and nothing clears it. With the `getApplicationManifests` calls on lines 519 and 542 deleted, `go test -run TestGetApplicationManifestsAppNamespaceArg ./internal/argocd/` still passes; a package-level `addNotice = config.AddNotice` seam (like `execArgoCdCli`) would let each subtest capture only its own notices.

internal/config/context.md
```diff
@@ -39,9 +39,10 @@ There are two producers, and `Notices()` returns them in this order:
    deduped so a per-application code path can call it unconditionally, and capped at `maxNotices`
    (10) so it can't crowd the diffs out of the comment budget.
 
-Nothing calls `AddNotice()` yet — it is the seam for that first version-gated feature.
-`ConnectivityCheck()` in `internal/argocd/helper.go` already parses the client and server versions
-and then discards them, which is the obvious place to hook one in.
+The first caller is `supportsManifestsAppNamespace()` in `internal/argocd/argocd_client.go`: it
+raises a notice on both its false branches (client version undetectable, client older than
+`appNamespaceManifestsMinVersion`), so the PR comment explains why app-of-apps children outside
+ArgoCD's own namespace are missing rather than leaving the reader to guess.
```
vrivellino: bug: This inverts which Applications go missing: `--app-namespace` is the namespace of the parent passed to `argocd app manifests`, so an old CLI still finds children in any namespace under a parent in ArgoCD's own namespace. What it loses is every child of a parent outside that namespace.
```suggestion
`appNamespaceManifestsMinVersion`), so the PR comment explains why the children of app-of-apps
parents outside ArgoCD's own namespace are missing rather than leaving the reader to guess.
```

internal/argocd/helper_test.go
```diff
@@ -840,6 +840,74 @@ func TestArgoAppsWithChangesNamespacePropagation(t *testing.T) {
...
+// it can't tell them apart.
+func TestArgoAppsWithChangesDisambiguatesByNamespace(t *testing.T) {
+	resetSupportsManifestsAppNamespaceCache(t)
+	orig := execArgoCdCli
+	defer func() { execArgoCdCli = orig }()
+
+	const repoURL = "https://github.com/acme/widgets.git"
+	manifestA := argoAppManifestFixture("web", "team-a", repoURL)
+	manifestB := argoAppManifestFixture("web", "team-b", repoURL)
+	combined := append(append([]byte{}, manifestA...), append([]byte("\n---\n"), manifestB...)...)
```
vrivellino: nit: `slices.Concat` does this in one call, and it is already how `execArgoCdCli` joins argv.
```suggestion
	combined := slices.Concat(manifestA, []byte("\n---\n"), manifestB)
```

vrivellino (commented):
No blocking findings. The two new notice subtests do not test their own calls, and one doc line in `internal/config/context.md` names the wrong missing Applications.

internal/argocd/argocd_client_test.go
brushmtn: Fixed in cbd5a9e: added a package-level `addNotice = config.AddNotice` seam (same pattern as `execArgoCdCli`), and both subtests now swap it to capture only their own call's notices instead of reading `config.Notices()`. Verified your repro (deleting the `getApplicationManifests()` calls) now fails, then restored.

internal/config/context.md
brushmtn: Applied your suggestion verbatim in cbd5a9e.

internal/argocd/helper_test.go
brushmtn: Applied your suggestion verbatim in cbd5a9e.

internal/argocd/context.md
```diff
@@ -59,10 +59,20 @@ resolves `argocd version --client` (client-only, **no server round-trip** — no
 `ConnectivityCheck()`) and caches a successful result process-wide via `cachedSupportsManifestsAppNamespace`
 (a failed lookup is *not* cached, so it's retried on the next call). If the pinned CLI is `<3.5.0`,
 the flag is omitted entirely and app-of-apps discovery falls back to the CLI's implicit default
-namespace — nested Applications outside that namespace won't be found on an old CLI.
+namespace — nested Applications outside that namespace won't be found on an old CLI. Both false
```
vrivellino: bug: This line has the same inversion that `cbd5a9e` fixed in `internal/config/context.md`. An old CLI still finds children in any namespace under a parent in ArgoCD's own namespace; what it loses is every child of a parent outside that namespace.
```suggestion
namespace — an old CLI can't find the children of a parent outside that namespace. Both false
```

internal/argocd/context.md
```diff
@@ -59,10 +59,20 @@ resolves `argocd version --client` (client-only, **no server round-trip** — no
...
+branches of `supportsManifestsAppNamespace()` (version undetectable, version too old) call
+`config.AddNotice()`, so the PR comment tells the reader a pinned CLI is why children are missing,
+rather than leaving them to read it off an unrelated ArgoCD error.
```
vrivellino: convention: This file does not document the new `addNotice` seam, and line 23 still calls `execArgoCdCli` "the mocking seam for this whole package". A later notice test that follows this file will read `config.Notices()` and pass without its own call, which is the bug `cbd5a9e` fixed.
```suggestion
`config.AddNotice()`, so the PR comment tells the reader a pinned CLI is why children are missing,
rather than leaving them to read it off an unrelated ArgoCD error. They call it through `addNotice`,
a second package-level seam: tests swap it to capture only their own call's notices, because
`config.Notices()` reads a process-wide list that tests in this package cannot clear.
```

vrivellino (commented):
No blocking findings. The 3 earlier findings are fixed in `cbd5a9e`, and both notice subtests now fail when their own `getApplicationManifests` call is deleted. Two doc gaps remain in `internal/argocd/context.md`.

internal/argocd/argocd_client.go
```diff
@@ -217,12 +222,12 @@ func supportsManifestsAppNamespace(ctx context.Context) bool {
 	v, err := argocdClientVersion(ctx)
 	if err != nil {
 		log.Warn().Err(err).Msg("failed to determine argocd CLI client version; omitting --app-namespace from `argocd app manifests` (app-of-apps discovery will only find nested apps in the CLI's default namespace)")
```
vrivellino: bug: This warning has the same inversion fixed in `internal/config/context.md`: an old CLI still finds nested apps in any namespace under a parent in the CLI's default namespace. What it loses is every child of a parent outside that namespace, so an operator who reads this log looks in the wrong place.
```suggestion
		log.Warn().Err(err).Msg("failed to determine argocd CLI client version; omitting --app-namespace from `argocd app manifests` (app-of-apps discovery will not find the nested apps of a parent outside the CLI's default namespace)")
```

vrivellino (commented):
One more doc gap, in an operator-facing log message.

internal/argocd/argocd_client.go
```diff
@@ -217,12 +222,12 @@ func supportsManifestsAppNamespace(ctx context.Context) bool {
...
-		config.AddNotice("argo-diff could not determine the `argocd` CLI version, so app-of-apps children are only enumerated for Applications in ArgoCD's own namespace.")
+		addNotice("argo-diff could not determine the `argocd` CLI version, so app-of-apps children are only enumerated for Applications in ArgoCD's own namespace.")
 		return false
 	}
 	supported := versionAtLeast(v, appNamespaceManifestsMinVersion)
 	if !supported {
-		config.AddNotice(fmt.Sprintf(
+		addNotice(fmt.Sprintf(
 			"The pinned `argocd` CLI (%s) is older than %s, so app-of-apps children are only enumerated for Applications in ArgoCD's own namespace. Point `ARGOCD_CLI_CMD_NAME` at a newer CLI to diff nested Applications in other namespaces.",
```
vrivellino: bug: "to diff nested Applications in other namespaces" tells the PR reader that a newer CLI finds children that live in other namespaces, but an old CLI already finds those under a parent in ArgoCD's own namespace. What a newer CLI adds is the nested Applications of parents in other namespaces.
```suggestion
			"The pinned `argocd` CLI (%s) is older than %s, so app-of-apps children are only enumerated for Applications in ArgoCD's own namespace. Point `ARGOCD_CLI_CMD_NAME` at a newer CLI to diff the nested Applications of parents in other namespaces.",
```

vrivellino (commented):
One more wording fix, in the old-client PR comment notice.

internal/argocd/context.md
brushmtn: Applied your suggestion verbatim in f5a6ab0.

internal/argocd/context.md
brushmtn: Applied your suggestion verbatim in f5a6ab0, and also updated the execArgoCdCli line at the top of the file to point at addNotice as the second seam.

internal/argocd/argocd_client.go
brushmtn: Applied your suggestion verbatim in f5a6ab0.

internal/argocd/argocd_client.go
brushmtn: Applied your suggestion verbatim in f5a6ab0.

vrivellino (commented):
No blocking findings. All 4 comments from the last round are fixed in `f5a6ab0`.

- nit: `internal/argocd/argocd_client.go:209-210` (outside the diff hunks) says that without the version check only discovery "for apps outside the CLI's default namespace" would pass the unknown flag. `getApplicationManifests()` passes `--app-namespace` for every parent, including parents in `argocd`, so an old CLI would reject every `argocd app manifests` call; suggested wording: "without this check, every `argocd app manifests` call would pass a flag the CLI doesn't understand."

brushmtn:
Applied the suggested wording in eb4c51e (`argocd_client.go:208-210`).

vrivellino (approved):

PR #327, opened 2026-09-15
fix(test): read pod logs once in run-argo-diff-pod.sh

brushmtn:
The `REQUIRE_LOG` assertion in `test/run-argo-diff-pod.sh` piped `kubectl logs` into `grep -q` while the script runs under `set -o pipefail`. `grep -q` exits at the first match, kubectl then takes EPIPE writing the rest of the stream and dies of SIGPIPE — silently, with status 141, since that is Go's runtime default for EPIPE on stdout — and pipefail turns that into a failed assertion even though the substring was present. This fix reads the pod logs once into a variable, prints that copy, and matches it with a here-string, so no pipeline is involved and the second `kubectl logs` call goes away.

It is a race on stream chunk boundaries rather than a content problem. [Run 34912743532](https://github.com/vince-riv/argo-diff/actions/runs/34912743532) failed this way: the pod log it printed *did* contain `Failed to diff application broken-deployment`, the end-to-end scenario behaved exactly as intended, and the log is byte-identical (19644 vs 19643 bytes) to the one from passing run 34850983335. The matched line sits ~380 bytes from the end of a ~19 KB log, so kubectl usually still has data to write when grep leaves.

No dependency changes. `test/context.md` gains a note so the pipeline form does not come back.


brushmtn:

<details>
<summary>Plan</summary>

1. Reproduce the mechanic locally (`pipefail` + early `grep -q` exit gives pipeline status 141).
2. Confirm the failing and passing e2e runs produced identical pod logs, ruling out a content change.
3. Read the pod logs once into a variable in `test/run-argo-diff-pod.sh`; print that copy.
4. Replace the `kubectl logs | grep -qF` assertion with a here-string match against the variable.
5. Record the gotcha in `test/context.md`.
6. Syntax-check the script, commit, push, open the PR.

</details>

<details>

**Status: complete, awaiting review.**

Done and verified:

- Root cause confirmed. `set -o pipefail` plus `grep -q` exiting early makes `kubectl` die of SIGPIPE (exit 141), which fails the `REQUIRE_LOG` assertion despite a match. Reproduced locally with a chunked writer: pipeline status 141 with `pipefail`, 0 without it.
- Evidence the e2e run itself was fine: run 34912743532's printed pod log contains `Failed to diff application broken-deployment`, and that log is 19644 bytes against 19643 bytes for passing run 34850983335 — effectively identical.
- `test/run-argo-diff-pod.sh` now captures logs once (`pod_logs=$(kubectl -n argocd logs "$pod_name" 2>&1 || true)`), prints them with `printf`, and asserts with `grep -qF -- "$REQUIRE_LOG" <<<"$pod_logs"`.
- `bash -n test/run-argo-diff-pod.sh` passes. `shellcheck` is not installed in this environment, so it was not run.
- `test/context.md` documents why the pipeline form must not return.

Not done: the script only runs inside the k3s end-to-end job, so it cannot be executed locally. Real validation is the `K3s Argo-Diff Test` workflow on this PR (it triggers on `test/**`). If that job's "Run argo-diff (manifest failure)" step passes, the change is confirmed.

Note for a fresh agent: `2>&1` on the `kubectl logs` capture is deliberate. It keeps any kubectl error text visible in the printed dump, which previously went to the step log directly.

</details>

vrivellino (approved):

PR #328, opened 2026-09-15
docs: clarify ARGO_DIFF_REQUIRE_APP_MATCH has no effect under GitHub Actions

brushmtn:
The `ARGO_DIFF_REQUIRE_APP_MATCH` row added in #315 claims GitHub Actions mode "honors this flag too". It does not, and it should not. `server.eventInfoFromEnv()` sets `EventInfo.Refresh` unconditionally so the action can re-read the PR's sha and refs, and the match check in `ProcessCodeChange()` sits behind `!eventInfo.Refresh` — so the flag is inert under Actions. The same row's GitHub Actions input column already said `N/A`, and `action.yml` defines no such input, so the prose contradicted both the table and the code.

This was raised in review on #315 ([discussion_r4005225849](https://github.com/vince-riv/argo-diff/pull/315#discussion_r4005225849)) as a bug to fix by splitting `Refresh` into two fields. It isn't one. The flag exists to stop argo-diff commit statuses landing on every PR of an org-wide webhook install; Actions mode skips commit statuses entirely and only runs in repos someone deliberately added the action to, so there is no chatter there to suppress. Being inert under Actions is the correct behavior — so this documents it instead of adding a field to support a combination nobody wants.

Note that Actions is the *only* mode the flag is inert in. `-f` event-file mode decodes `Refresh` from the event JSON, whose documented example omits the key, so the check runs there just as it does for webhook events. The wording throughout says "no effect under GitHub Actions" rather than "webhook-only" for that reason.

Changes: corrected the README row; expanded `internal/process_event/context.md` step 4 to spell out that the `!eventInfo.Refresh` bypass covers two distinct cases and that `-f` mode is not a third; added the same explanation as a code comment at the bypass, since it named only one of the two reasons and that is what made this look like an oversight; exported `requireAppMatch()` → `RequireAppMatch()` so `cmd/main.go` warns when an operator sets the flag in an Actions runner anyway; and documented that warning plus the resulting import in `cmd/context.md`. Exporting keeps one source of truth for the parsing rules rather than duplicating the trim/case-fold in main.

No behavior change in any mode. No new dependencies.


brushmtn:

<details>
<summary>Plan</summary>

1. Correct the `ARGO_DIFF_REQUIRE_APP_MATCH` row in `README.md` to state the flag is webhook-only, mirroring the existing `ARGO_DIFF_REFRESH_COMMENT_KEYWORDS` row's wording.
2. Expand step 4 in `internal/process_event/context.md` to explain that `EventInfo.Refresh` carries two meanings, and that Actions mode setting it unconditionally makes the flag inert there by design.
3. Add a code comment at the `!eventInfo.Refresh` bypass in `code_change.go` naming both reasons for the bypass.
4. Export `requireAppMatch()` as `RequireAppMatch()`, and have `cmd/main.go` log a warning when the flag is set under `GITHUB_ACTIONS=true`.
5. Verify: `go build ./...`, `gofmt -l .`, `go test ./...`, `golangci-lint run`.

</details>

<details>

**Complete, and the first review round is addressed.** Commits: d1c39ab (initial), 30b333c (review fixes).

Verified after the latest commit: `go build ./...` clean; `gofmt -l .` clean; `go test ./...` fully passing; `golangci-lint run` reports 0 issues.

Review round 1 (vrivellino, no blocking findings) raised two points, both fixed in 30b333c. First, the "Only takes effect when deployed as a webhook service" wording was wrong for `-f` event-file mode: `server.eventInfoFromFile()` plain-decodes `EventInfo`, and the README's example event JSON omits the `refresh` key, so `Refresh` is false and the match check does run there. Every description is now reframed as "inert under GitHub Actions only", which is the claim the code supports — this covered the README row, the `RequireAppMatch()` doc comment, and the `cmd/main.go` warning string. Second, `cmd/context.md` step 5 now documents the Actions-mode warning and the `internal/process_event` import it required.

Decision worth recording for anyone revisiting this: review on #315 proposed splitting `EventInfo.Refresh` into two fields ("re-read PR metadata" vs "a human asked for this run") because Actions mode sets it unconditionally and so bypasses the match check. That split was deliberately **not** done. The flag's only purpose is suppressing commit-status noise on org-wide webhook installs, and Actions mode skips commit statuses anyway, so there is no case where someone wants the flag active under Actions. The fix is documentation plus a warning, not a second field.

No follow-up work outstanding.

</details>

README.md
```diff
@@ -264,7 +264,7 @@ can still be set through the chart's `deployment.env` / `deployment.envFrom` pas
...
+| ARGO_DIFF_REQUIRE_APP_MATCH  | N/A                         | no               | `false`  | Set to `true` to check for a matching ArgoCD application *before* posting anything to GitHub, and stay completely silent (no commit status, no comment) when nothing matches, instead of posting a `pending` status up front followed by an unconditional `success`. Mainly useful when deployed as a webhook server installed org-wide: without it, every PR in every repo the GitHub App can see gets an argo-diff commit status, even repos ArgoCD doesn't track, which tends to confuse people. The match check itself is cheap (one `argocd app list` call plus local filtering, no `argocd app diff` calls), but it does mean one extra `argocd app list` call on PRs that do match, and it skips clearing a stale prior comment if a later push makes a previously-matching PR stop matching. Only takes effect when deployed as a webhook service — GitHub Actions mode skips commit statuses already and only runs in repos the action was added to, so there is no chatter to suppress; argo-diff logs a warning if the flag is set there. |
```
vrivellino: bug: "Only takes effect when deployed as a webhook service" is wrong for `-f` event-file mode: the README's own example event JSON has no `refresh` key, so `Refresh` decodes to `false` and the match check runs. A developer who sets the flag and runs `go run cmd/main.go -f event.json` against an untracked repo gets a silent no-op; the `cmd/main.go` warning text ("only applies to webhook server deployments") has the same gap.

```suggestion
| ARGO_DIFF_REQUIRE_APP_MATCH  | N/A                         | no               | `false`  | Set to `true` to check for a matching ArgoCD application *before* posting anything to GitHub, and stay completely silent (no commit status, no comment) when nothing matches, instead of posting a `pending` status up front followed by an unconditional `success`. Mainly useful when deployed as a webhook server installed org-wide: without it, every PR in every repo the GitHub App can see gets an argo-diff commit status, even repos ArgoCD doesn't track, which tends to confuse people. The match check itself is cheap (one `argocd app list` call plus local filtering, no `argocd app diff` calls), but it does mean one extra `argocd app list` call on PRs that do match, and it skips clearing a stale prior comment if a later push makes a previously-matching PR stop matching. Has no effect in GitHub Actions mode — it skips commit statuses already and only runs in repos the action was added to, so there is no chatter to suppress; argo-diff logs a warning if the flag is set there. |
```

cmd/main.go
```diff
@@ -110,6 +111,9 @@ func main() {
 	// if running under Github Actions, skip github connectivity check
 	if os.Getenv("GITHUB_ACTIONS") == "true" {
 		log.Info().Msg("GITHUB_ACTIONS set in the environemtn - running once with event data from environment")
+		if process_event.RequireAppMatch() {
```
vrivellino, line 114: convention: AGENTS.md requires updating a directory's `context.md` when its behavior changes, and `cmd/context.md` step 5 still describes the Actions branch without this warning. Add a line there so the `process_event` import and warning are documented.

vrivellino (commented):
No blocking findings. All checks pass.

README.md
brushmtn: Good catch — fixed in 30b333c, took your suggestion for the row. Confirmed the gap: `eventInfoFromFile()` plain-decodes `EventInfo`, the example event JSON at README:319 has no `refresh` key, so `Refresh` is false and the check runs under `-f`.

Reframed everything from "webhook-only" to "inert under Actions only", which is the claim the code actually supports. That also covers the `cmd/main.go` warning text and the `RequireAppMatch()` doc comment, which both had the same gap, plus a note in `internal/process_event/context.md` so `-f` mode is not mistaken for a third bypass case.

cmd/main.go
brushmtn: Fixed in 30b333c. `cmd/context.md` step 5 now documents the warning, why the flag is inert under Actions, and that this is the only reason `cmd` imports `internal/process_event`, with a pointer to `internal/process_event/context.md` step 4.

vrivellino (approved):

